PR #5, opened 2026-05-26
feat: expose /metrics endpoint for Prometheus scraping

turfin-logic:
Closes #3

Hi @SikamikanikoBG! Here is the complete implementation as discussed in issue #3.

### `app.py`
- Added `prometheus_client` import (with graceful fallback if not installed)
- Defined **11 Gauges** at module level (defined once, never re-created):
  - `homelab_gpu_vram_used_mb`, `homelab_gpu_vram_total_mb`, `homelab_gpu_util_pct`, `homelab_gpu_temp_c`, `homelab_gpu_power_w`
  - `homelab_host_cpu_pct`, `homelab_host_mem_used_pct`, `homelab_host_disk_used_pct`
  - `homelab_container_state`, `homelab_systemd_unit_state`
  - `homelab_model_loaded_vram_mb`
- Added `GET /metrics` route that reads **exclusively from `LATEST` and `HEALTH` in-memory snapshots** — exactly as you requested: the exporter is a pure translation layer with zero extra I/O and zero double-sampling.

### `Dockerfile`
- Added `prometheus_client==0.20.0` to the single `pip install` line (keeps the image slim).

### `docs/grafana/homelab_prometheus_dashboard.json`
- Ready-to-import Grafana dashboard covering GPU VRAM, utilisation, temperature, host CPU/RAM, disk usage per mountpoint, and model VRAM loaded.
- Import via **Grafana → Dashboards → Import → Upload JSON file**.

### `README.md`
- New **Prometheus integration** section with: full metrics table, `curl` verification step, sample `prometheus.yml` scrape config, and Grafana import instructions.

## Quick verification

```bash
docker compose up -d --build
curl http://localhost:9800/metrics
```

Expected output:
```
# HELP homelab_gpu_vram_used_mb GPU VRAM used (MB)
# TYPE homelab_gpu_vram_used_mb gauge
homelab_gpu_vram_used_mb{gpu="gpu0"} 4096.0
...
```

Let me know if you'd like any adjustments!

PR #7, opened 2026-05-26
fix: clear stale label series on each /metrics scrape

### LINKED ISSUE #6: Prometheus exporter: clear stale label series on each scrape
Follow-up to #5 (thanks @turfin-logic — the exporter works great).

**Problem:** the labelled gauges (`homelab_container_state`, `homelab_systemd_unit_state`, `homelab_model_loaded_vram_mb`, `homelab_host_disk_used_pct`) are `.set()` on each scrape but never cleared, so old label combinations linger at their last value for the life of the process. Examples:

- A container that goes `running → exited` keeps reporting `1` for **both** states (the old `state="running"` series is never updated again).
- A systemd unit that goes `active → failed` keeps its old `state="active"` series at `1`.
- An unloaded model / removed mountpoint keeps reporting its last value.

Not dangerous — just misleading metrics over long uptime.

**Suggested fix:** call `.clear()` on each labelled gauge at the top of the `/metrics` handler before re-setting, e.g. `_G['container_state'].clear()`. (Single-label-set gauges like `host_cpu` don't need it.) Alternatively, implement a custom `Collector` that yields fresh series each scrape.

Good first issue — small, self-contained, no collector changes.

turfin-logic:
Closes #6

Hi @SikamikanikoBG! Spotted the same issue when reviewing the scrape behaviour — great catch. Here is the fix.

## What changed

**`app.py` — `/metrics` handler only**

Added a single `for` loop that calls `.clear()` on every multi-label gauge **before** re-populating it on each scrape:

```python
for key in ("gpu_vram_used", "gpu_vram_total", "gpu_util", "gpu_temp", "gpu_power",
            "host_disk_used", "model_vram", "container_state", "systemd_unit"):
    _G[key].clear()
```

This ensures that stale label combinations — a container that went from `running` → `exited`, an unmounted disk, an unloaded model, or a removed systemd unit — are not reported on the next scrape.

Single-label gauges (`host_cpu`, `host_mem_used`) are intentionally left alone: they have no label combinations to go stale and `.clear()` on them would cause a momentary gap in the series.

## Why this approach

`.clear()` is the idiomatic solution recommended by the `prometheus_client` maintainers for exactly this pattern (dynamic label sets that change at runtime). The alternative — a custom `Collector` — is more powerful but adds complexity that is not needed here.

## Verification

```bash
# Start the container, scrape once to populate series
curl -s http://localhost:9800/metrics | grep homelab_container_state

# Stop a container, scrape again — its label series should be gone
docker stop <some-container>
curl -s http://localhost:9800/metrics | grep homelab_container_state
# The stopped container's old state="running" line should no longer appear
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -583,6 +583,11 @@ def metrics():
     if not _PROM_OK:
         return Response("# prometheus_client not installed\n", mimetype="text/plain", status=503)
 
+    # Clear all multi-label gauges before re-populating so stale series vanish.
+    for key in ("gpu_vram_used", "gpu_vram_total", "gpu_util", "gpu_temp", "gpu_power",
+                "host_disk_used", "model_vram", "container_state", "systemd_unit"):
+        _G[key].clear()
+
     # ── GPU ──────────────────────────────────────────────────────────────────
     gpu_label = "gpu0"
     _G["gpu_vram_used"].labels(gpu=gpu_label).set(LATEST.get("mem_used", 0))


SikamikanikoBG (commented):
Great catch, and the fix itself is spot-on, @turfin-logic — `.clear()` on the multi-label gauges before re-populating is exactly the idiomatic fix, and leaving the unlabeled `host_cpu`/`host_mem_used` alone is the right call. 🙌

One small git gotcha to sort before I can merge: this branch was cut from `main` **before #5 merged**, so it carries all of the (now already-merged) Prometheus work again, which makes it show as *conflicting* and, more importantly, its net effect would **revert the version 0.4.0 → 0.3.1** in `app.py` and the README badge. Totally normal stale-branch thing — here's the ~30-second fix to reduce this PR down to just your `.clear()` loop:

```bash
# add the canonical repo as 'upstream' (skip if you already have it)
git remote add upstream https://github.com/SikamikanikoBG/homelab-monitor.git 2>/dev/null; git fetch upstream

# reset your branch onto the latest main (drops the already-merged #5 work)
git checkout fix/clear-stale-label-series
git reset --hard upstream/main
```

Then re-add just the fix in `app.py`, right before the `# ── GPU ──` block in `metrics()`:

```python
    # Clear all multi-label gauges before re-populating so stale series vanish.
    for key in ("gpu_vram_used", "gpu_vram_total", "gpu_util", "gpu_temp", "gpu_power",
                "host_disk_used", "model_vram", "container_state", "systemd_unit"):
        _G[key].clear()
```

```bash
git add app.py
git commit -m "fix: clear stale label series on each /metrics scrape"
git push --force-with-lease
```

That updates this same PR in place — it'll become a clean ~6-line diff with no conflict. No need to touch the version; I'll bump it on merge. Ping me once it's pushed and I'll merge right away. Thanks again for the quick follow-up! 🚀

turfin-logic:
Rebased on \upstream/main\ as requested to drop the already-merged #5 commits. Just a clean 6-line diff now! Ready for merge! 🚀

SikamikanikoBG (approved):
Perfect — clean as a whistle now, just the `.clear()` loop on top of current main. Thanks for the quick rebase turnaround, @turfin-logic! Merging. 🚀

Issue #32, opened 2026-05-29
Mini top — show top-N processes by CPU/RAM on the Host tab


SikamikanikoBG:
## What
A compact two-column table on the **Host** tab showing the **top 10 processes by CPU** and **top 10 by RAM**, refreshed every 15s like the rest of the host KPIs.

## Why
When the box is grumpy, you immediately want to know *who's grumpy*. Right now that means an ssh + `top`. With this, no ssh, just glance.

### Backend (`app.py`)
- `collect_top_processes()` reads `/proc/[pid]/stat` + `/proc/[pid]/status` for all PIDs (accessible because of `pid: host`).
- Aggregate by executable command (so 20 `python` workers collapse into one row), sort by CPU% and by RSS, take top 10 each.
- Expose via `/api/data` or a new `/api/processes`.

### Frontend (`static/dashboard.html`)
- A small "🧠 Top processes" card on the Host tab with two columns: *By CPU* / *By RAM*.

## Acceptance
- [ ] Each column lists ~10 processes with sensible numbers.
- [ ] Refreshes on the same cadence as other host KPIs.

## Pointers
- `/proc` is already mounted; example `psutil`-free `/proc` parsing patterns are easy to find.
- Don't add psutil as a dep — keep it lean, the existing host collector reads `/proc` directly.

🟢 good first issue · 🐧 linux internals · 🧠 mini-htop

samkelomncwabe63-svg:
Hi there! I'm a first-time contributor and I'd love to take this on if it's still available.

SikamikanikoBG:
Welcome aboard! Please fork and develop the feature, submit PR and I will review it.

SikamikanikoBG:
Heads up for anyone eyeing this: there's already a draft PR in progress for it — #34 by @samkelomncwabe63-svg. The collector's mostly there; it just needs the frontend tab to land. If you'd like to help, jump in on the PR rather than starting fresh 🙌

PR #34, opened 2026-05-29
feat: add lightweight parsing for top memory processes

samkelomncwabe63-svg:
## Checklist

- [ ] Branched off the **latest `main`** (`git pull origin main` before
      behaves as expected.
- [ ] No version bump in this PR — version bumps are handled by the maintainer
      at release time.

### DIFF
--- a/app.py
+++ b/app.py
@@ -460,9 +460,39 @@ def collect_update():
         _UPDATE_CACHE.update({"at": now, "data": data})
     return data
 
+
+def collect_processes():
+    import os
+    processes = []
+    try:
+        pids = [p for p in os.listdir('/proc') if p.isdigit()]
+        for pid in pids:
+            try:
+                with open(f'/proc/{pid}/stat', 'r') as f:
+                    stat = f.read().split()
+                name = stat[1].strip('()')
+                
+                mem = 0
+                with open(f'/proc/{pid}/status', 'r') as f:
+                    for line in f:
+                        if line.startswith('VmRSS:'):
+                            mem = int(line.split()[1]) // 1024
+                            break
+                
+                if mem > 0:
+                    processes.append({"pid": pid, "name": name, "mem_mb": mem})
+            except (IOError, IndexError):
+                continue
+    except Exception:
+        return []
+
+    top_mem = sorted(processes, key=lambda x: x['mem_mb'], reverse=True)[:10]
+    return {"top_memory": top_mem}
+
 def health_scan():
     HEALTH["docker"]  = collect_docker()
     HEALTH["systemd"] = collect_systemd()
+    HEALTH["processes"] = collect_processes()
     HEALTH["update"]  = collect_update()
     HEALTH["at"]      = int(time.time())
 


SikamikanikoBG (commented):
Thanks for the PR @samkelomncwabe63-svg! 👋 Top-processes is a nice addition — the collector is most of the way there. A few things needed before we can land it:

### Match the "add a monitor" pattern from `CONTRIBUTING.md`

The project's convention is collector → `health_scan()` → `/api/health` → **TABS entry + section + renderer**, with collectors returning a standard shape so the dashboard can degrade gracefully:

```python
{"available": bool, "summary": {...}, "items": [...]}
```

Two things missing here:

1. **Frontend tab.** Right now the data is collected but nothing displays it. Could you add the three small pieces in `static/dashboard.html`?
   - one entry to the `TABS` array
   - one `<section data-tab="...">` for the tab body
   - a small renderer that reads from `/api/health`

   No build step or framework — just vanilla JS like the other tabs.

2. **Return shape.** Currently `collect_processes()` returns `{"top_memory": [...]}` on success and a bare `[]` on failure. Could you switch to:
   - on Linux with data: `{"available": True, "summary": {...}, "items": [...]}`
   - on non-Linux or on error: `{"available": False, "summary": {}, "items": []}`

   That way the renderer can show a clean "not supported on this host" state instead of erroring, and it matches what the other collectors do.

### Smaller things while you're in there

- **`/proc/<pid>/stat` truncates `comm` to 15 bytes**, so long-named processes (e.g. `qemu-system-x86_64`) show up wrong. `/proc/<pid>/comm` or `/proc/<pid>/cmdline` (with a small truncation guard) reads the full name.
- **`import os` inside the function** can go — it's already imported at the top of `app.py`.
- For the PR itself: please branch off `main` rather than pushing from your fork's `main` (CONTRIBUTING.md has the recipe), and a one-line description + a screenshot of the new tab would help future reviewers.

Happy to take another pass once those are in — thanks again for jumping in! 🛰️

SikamikanikoBG:
Hey @samkelomncwabe63-svg — checking back in 👋 This one's still very much wanted: it maps straight to #32 (top-N processes on the Host tab), which is open and unclaimed — so there's a clear path to merge.

Nothing's changed on my side; the review above still stands as the checklist. The single highest-leverage next step is the **frontend tab** — the three small vanilla-JS pieces in `dashboard.html`. The collector's already most of the way there.

If you've got time for another pass, I'm happy to unblock anything — and if you'd rather pair on the frontend bit, just say the word. Would genuinely love to land this as your first merged one. 🛰️

SikamikanikoBG:
Thank you for this — it actually seeded the Top-processes feature in the dashboard. That work has since been built out on `next` as `collect_top_processes()`: it reports both CPU% and RAM, aggregates by command (so multi-worker apps collapse into one row), parses `comm` robustly, and returns cleanly on hosts without `/proc` (e.g. Windows hubs). You're credited by handle in the `app.py` comment above that function.

Because the feature now ships in a superset form, there's nothing left to merge here — the branch is also several releases stale against `next`. I'd suggest closing this as already incorporated. Really appreciate the contribution that kicked it off!

SikamikanikoBG:
Closing as incorporated — the Top-processes feature this seeded now ships on `next` as `collect_top_processes()`, and you're credited by handle in the code. Thank you, @samkelomncwabe63-svg — really appreciate the contribution that kicked it off.

PR #88, opened 2026-06-10
fix: refresh Services tab on host switch

### LINKED ISSUE #82: Services tab doesn't refresh when you switch hosts — you have to tab away and back
## What
On the **Services** tab, switching the active host doesn't re-render the list. The page keeps showing the *previous* host's services. The only way to see the newly-selected host's services is to click another tab (e.g. **Containers**) and then click back to **Services**.

## Why
It looks broken: you pick a host, nothing changes, and you're left wondering whether the switch even registered (or worse, whether you're reading the wrong machine's data). Every other tab updates on host-switch — Services is the odd one out.

## Likely cause
The host-selector's change handler re-renders the *currently active* tab for most tabs, but the Services renderer is only wired to run on **tab-change**, not on **host-change**. So a host switch updates the in-memory selection but never repaints the Services section until the next tab activation.

## Scope (frontend)
- In the host-switch handler, if the active tab is **Services**, re-fetch/re-render it immediately — same path a tab-change takes.
- Sanity-check the other tabs while in there: anything else that only repaints on tab-change should repaint on host-change too.

## Acceptance
- [ ] Switch host while sitting on the **Services** tab → the list updates immediately, no tab-dance.
- [ ] No stale rows from the previously-selected host linger.

🐛 bug · 🖱️ one-handler fix · 🟢 good first issue

mohd-ibadullah:
## What this changes

Fixes #82 by refreshing the Services tab when the selected host changes.

When a host is switched while the Services tab is active, `refreshCurrentTab()` now calls `renderServicesTab()` so the services list is re-rendered immediately.

Fixes #82

## Checklist

* [x] Base branch is `next`
* [x] No version bump included
* [x] Minimal scoped fix

## Notes for the reviewer

This is a one-line fix that reuses the existing `renderServicesTab()` render path and only affects host-switch refresh behavior.


SikamikanikoBG:
Thanks for this, @mohd-ibadullah — clean, well-scoped fix. The missing services branch in `refreshCurrentTab()` was exactly the gap behind #82.

What I did on my end: reviewed, retargeted onto `next`, CI smoke passed green, and squash-merged — no version bump, so it'll ride along in the next release rather than a hotfix.

First contribution to the repo too — welcome aboard, and glad to have you. 🙌

PR #94, opened 2026-06-11
feat(services): detect ports from cgroup for forking services

### LINKED ISSUE #83: Show each service's port so you know what's on :8080 at a glance
## What
On the **Services** tab, show each service's **listening port** as a small badge/column next to its name — when we can detect one.

## Why
Right now a service like **Pi-hole** shows up with no port, so you can't tell at a glance what's actually on `:80` / `:53`, and you can't click through. The port is half of "what is this and how do I reach it?" — surfacing it turns the Services tab from a list of names into something you can actually navigate from.

### Backend
- For each detected service, resolve its listening port(s). The data is already there to be mined — e.g. parse `ss -ltnp` / `/proc/net/tcp` and map the owning PID → service, or read `ListenStream`/`ListenPort` from the systemd unit. Fall back to a well-known-port map for the usual suspects (Pi-hole, etc.).
- A service can listen on several ports — return a small list, not just one.

- Render the port(s) as a muted badge next to the service name. If the port maps to an HTTP service, make it a clickable `http://<host>:<port>` link.

## Acceptance
- [ ] Services with a detectable listening port show it (Pi-hole shows its port).
- [ ] Services with no detectable port simply show no badge — no "unknown"/"0" noise.
- [ ] HTTP-ish ports are click-through links.

🟢 good first issue · 🔌 quality-of-life · 🐧 linux internals

ravvdevv:
For Type=forking services (e.g. Pi-hole FTL, dnsmasq), the listening socket lives in a child process rather than MainPID.  Previously _ports_for_pid only checked MainPID, so these services showed no port.

Fix: read ControlGroup from systemd service props, walk cgroup.procs to get all PIDs in the unit, then union _ports_for_pid across them. Falls back gracefully if cgroup is unreadable.

- Add _cgroup_pids() helper in app.py
- Modify collect_systemd() to fall back to cgroup PIDs when MainPID has no listening ports
- No frontend changes needed — renderPorts() already handles display
- No logic changes to existing behavior for non-forking services

Fixes #83

## Checklist

- [ ] **Base branch is `next`** (the integration branch), not `main`. Change the
- [ ] Branched off the **latest `next`** (`git pull origin next` before
      behaves as expected.
- [ ] No version bump in this PR — version bumps are handled by the maintainer
      at release time.

## Notes for the reviewer
Anything worth flagging: trade-offs you considered, screenshots of UI
changes, things you're unsure about.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1292,23 +1292,34 @@ def _listen_inode_to_port():
 def _cgroup_pids(control_group):
     """Return all PIDs in a systemd service's cgroup.
 
-    Reads /sys/fs/cgroup/<control_group>/cgroup.procs to get every process
-    belonging to the unit. This covers child/forking processes that may own
-    listening sockets even though MainPID does not. Returns an empty list if
-    the cgroup path is empty or unreadable (e.g. on hosts without cgroupfs).
+    Instead of reading /sys/fs/cgroup/.../cgroup.procs (which is unavailable
+    inside Docker's private cgroup namespace), we scan /proc/<pid>/cgroup for
+    every running process and match the service's ControlGroup string. Works
+    on bare-metal and in Docker, and avoids cgroup v1/v2 path differences.
+
+    Falls back to an empty list if the cgroup path is empty or no PIDs match.
     """
     if not control_group:
         return []
-    # Normalize: systemd may return the cgroup path with a leading slash.
-    # cgroup v1 and v2 both expose cgroup.procs under the same hierarchy.
-    for base in ("/sys/fs/cgroup", "/sys/fs/cgroup/unified"):
-        path = f"{base.rstrip('/')}/{control_group.lstrip('/')}/cgroup.procs"
-        try:
-            with open(path) as f:
-                return [int(line.strip()) for line in f if line.strip()]
-        except (FileNotFoundError, PermissionError, ValueError, OSError):
-            continue
-    return []
+    # Normalize: ControlGroup may or may not have a leading slash; strip it
+    # so we can match against the proc entries consistently.
+    cg = control_group.lstrip("/")
+    pids = []
+    try:
+        for pid_dir in os.listdir("/proc"):
+            if not pid_dir.isdigit():
+                continue
+            try:
+                with open(f"/proc/{pid_dir}/cgroup") as f:
+                    for ln in f:
+                        if cg in ln.strip():
+                            pids.append(int(pid_dir))
+                            break
+            except (FileNotFoundError, PermissionError, OSError):
+                continue
+    except Exception:
+        return []
+    return pids
 
 def _ports_for_pid(pid, inode_to_port):
     """LISTEN ports owned by `pid` by walking /proc/<pid>/fd/* for socket:[N]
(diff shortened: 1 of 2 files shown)


SikamikanikoBG:
Really nice work, @ravvdevv — right direction and a clean diff. I tested it on a real install though, and there's one blocker plus two tweaks:

1. **It won't fire in the Docker deployment** — the main way people run this. Inside the container the cgroup namespace is private: `/proc/self/cgroup` is just `0::/` and `/sys/fs/cgroup/system.slice/<svc>/cgroup.procs` doesn't exist, so `_cgroup_pids()` returns `[]` and Pi-hole still shows no port. (The existing MainPID detection works in-container only because it goes through `/proc`, which is the host's thanks to `pid: host` — not through `/sys`.) Cleanest fix: get the unit's PIDs via `/proc` instead of `/sys/fs/cgroup` — scan `/proc/<pid>/cgroup` and match the service's `ControlGroup` string. Works on bare-metal *and* in Docker, and sidesteps the cgroup v1/v2 path differences too.

2. **Union the ports across all the unit's PIDs.** Right now it breaks on the first child with a port — but a forking service can listen on several ports across children (Pi-hole does DNS + web). Your description even says "union", so I think it's just a small slip.

3. **Base branch → `next`, not `main`** — that's the integration branch (see the PR template checklist).

Happy to pair on the `/proc/<pid>/cgroup` matching if it's useful. Thanks again — the diagnosis and the graceful-fallback instinct are spot on.

ravvdevv:
Thanks for the review! All three points addressed in the latest push:

1. **Docker compat** — replaced the  path-based lookup with  string matching. Scans every running PID and matches the service's  string, so it works on bare-metal and in Docker's private cgroup namespace. Also sidesteps cgroup v1/v2 path differences.

2. **Union fix** — changed the child PID loop from `break` on first match to collecting + deduplicating ports across all children. Pi-hole should now show both DNS (53) and web (80/443) ports.

3. **Base branch** — retargeted to `next`.

Let me know if anything else needs tweaking!

SikamikanikoBG:
Thanks for the fast turnaround, @ravvdevv — base is on `next`, and the union + `/proc` switch read right in the code.

I re-tested in the container though, and there's a catch: `/proc/<pid>/cgroup` is namespace-virtualised too, same as `/sys`. Inside the container, host PID 1's cgroup reads `0::/../../init.scope` (not `/system.slice/...`), and matching a real service's `ControlGroup` returns 0 PIDs — so on the Docker deployment it still comes up empty. Both reads go through the container's private cgroup namespace — that's the actual root cause.

The missing piece is a one-line deployment change, not code: `cgroupns: host` in the compose. I verified it — with `--cgroupns=host`, host PID 1 reads `0::/init.scope` and `grep -l system.slice /proc/*/cgroup` matches the real service PIDs. With that in place your code works exactly as written — it'd even make the simpler `/sys` version viable. Since it touches the published compose posture, I'm happy to fold `cgroupns: host` into `docker-compose.yml` + `docker-compose.dev.yml` when we merge — or you can add it here, your call. The `app.py` change is good on top of it.

One smaller thing for later: it re-scans all of `/proc` once per port-less service — on a busy host that's a lot of reads each cycle. Scanning `/proc` once into a `cgroup -> pids` map would make it O(procs) instead of O(services × procs). Not a blocker.

Really solid work chasing this down — the namespace behaviour is genuinely non-obvious.

SikamikanikoBG:
Tested end-to-end and it works. I ran your functions inside the container and `libvirtd.service` now surfaces its child-owned `:53` — the dnsmasq the MainPID doesn't own — where before the namespace fix it matched 0 PIDs. nginx unions 80/443 across MainPID + worker too.

The only missing piece was the deployment side, so I pushed one commit folding `cgroup: host` into `docker-compose.yml` (with a comment on why). Your `app.py` change is spot on. Merging now — thanks for the great work and the quick iterations.

ravvdevv:
Nice work getting this merged! The cgroup namespace rabbit hole was a good catch — `/proc/<pid>/cgroup` virtualization inside containers is one of those things that looks like it should work until it does not. Glad the `cgroupns: host` deployment fix sealed it.

Definitely onboard with the `cgroup → pids` map optimization for busy hosts — O(procs) vs O(services × procs) is worth it once the service count grows. Will pick that up next.

Thanks for the thorough review and the fast merge, @SikamikanikoBG 🤝

PR #142, opened 2026-06-14
feat: opt-in one-click self-update from the dashboard

pehota:
## What this changes

Adds an opt-in **"Update now"** button to the update modal. When clicked, the monitor pulls the latest release image, recreates itself, verifies it came back healthy, and rolls back to the previous image automatically if it doesn't. Live progress and logs are streamed in the UI so you can watch it happen without leaving the dashboard.

Closes #141

## How it works

A container can't recreate itself in-process (the process dies mid-swap), so a detached `docker:cli` helper container runs the project's own compose file:
- `MONITOR_IMAGE=<repo>:<version> docker compose pull` — pulls the immutable versioned tag, not the moving `:latest`
- `docker compose up -d --no-build` — recreates the service
- Health-gates on `GET /api/health` for ~60s; if the new container isn't healthy or reports the wrong version, it redeploys the **previous image digest** and reports `rolled_back`

Progress is written to `./data/update_state.json` + `./data/update.log` so the UI survives the restart and reconnects cleanly.

Off by default (`ALLOW_SELF_UPDATE=1` to enable). When disabled, the modal keeps the existing manual command. When enabled, the manual command is hidden (redundant).

## New surface

- `POST /api/update/app` — starts the update; 400 if disabled/no-update/not-a-compose-deploy, 409 if already running
- `GET /api/update/app/status` — returns state + log tail; readable across restarts since it just reads `./data`
- `update.self_update_enabled` added to `/api/health`
- `docker-compose.yml`: socket changed `:ro` → rw (needed to create the helper); `image` parameterised as `${MONITOR_IMAGE:-…:latest}` so rollback can pin the previous digest; commented `ALLOW_SELF_UPDATE` env line added

## How to test before merging

The setup below fakes the running version to 0.15.0 so the real v0.16.0 release shows as an available update, using a local registry as the pull target — no published image needed.

```bash
# Prerequisites: Docker, ports 9800 and 5000 free
git clone https://github.com/pehota/homelab-monitor.git -b feat/self-update
cd homelab-monitor
DEMO=/tmp/hlm-pr-test; mkdir -p $DEMO/data

# Build base image
docker build -t hlm-pr-base .

# Build two versioned images and push to a local registry
docker run -d -p 5000:5000 --name hlm-pr-registry registry:2
for v in 0.15.0 0.16.0; do
  printf "FROM hlm-pr-base\nRUN sed -i 's/VERSION      = \"0.17.0-next\"/VERSION      = \"$v\"/' /app/app.py\n" \
[...]

PR #144, opened 2026-06-14
fix: restore :ro socket default; move self-update rw into an override file

pehota:
## What this changes

Follow-up to #142, addressing the reviewer's note: the main `docker-compose.yml` should keep the docker socket `:ro` by default, and only self-update users should opt into read-write.

Restores `:ro` in `docker-compose.yml` and ships a new `docker-compose.self-update.yml` override file. Self-update users enable it with:

```bash
docker compose -f docker-compose.yml -f docker-compose.self-update.yml up -d
```

Plain monitoring deployments are entirely unaffected — they keep the read-only socket.

## Checklist

- [x] **Base branch is `next`**
- [x] Branched off the latest `next`
- [x] `docker compose -f docker-compose.yml -f docker-compose.self-update.yml config` parses cleanly
- [x] No version bump in this PR

### DIFF
--- a/docker-compose.self-update.yml
+++ b/docker-compose.self-update.yml
@@ -1,0 +1,16 @@
+# Opt-in self-update override — use alongside docker-compose.yml:
+#
+#   docker compose -f docker-compose.yml -f docker-compose.self-update.yml up -d
+#
+# This upgrades the docker socket from :ro to read-write (needed to create the
+# detached docker:cli helper that recreates this container) and enables the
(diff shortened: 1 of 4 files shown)


SikamikanikoBG:
merged 🙌 that was fast — thank you @pehota!

this is exactly what i had in mind: `:ro` stays the default so normal monitoring setups keep the read-only socket, and self-update folks just add the override file. clean and well documented. really appreciate you jumping on it so quick 🔥

PR #151, opened 2026-06-15
Fix #24: Per-service notification rules — pick which services alert you

### LINKED ISSUE #24: Per-service notification rules — pick which services alert you
## What
Today the notifier fires on **any** container or systemd unit going unhealthy / failed. Useful when you run a handful of services, noisy when you run thirty. This adds a small **routing layer** between detection and dispatch:

- A list of rules in the **Alerts** tab — *e.g. "if container `ollama-*` becomes unhealthy → Discord, min severity `warning`"*.
- Default rule (matches everything → both channels) stays as today, so plug-and-play is preserved.
- Wildcards on the service name (`postgres-*`, `*-prod`), per-channel and per-minimum-severity filtering.

## Why
- Users opt **in** to alerts they care about, out of ones they don't.
- Different channels for different concerns ("look at this when you can" on Discord vs "drop everything" on ntfy).
- Existing edge-trigger / no-spam guarantees from v0.5.0 stay untouched — rules only decide *whether* a fire happens, not when.

### Backend (`app.py`)
- New `notification_rules` table — `(id, match_kind, match_pattern, channel, min_level, enabled)`.
- `match_kind`: `container | systemd | gpu | host` (the 4 alert origins we already have).
- `match_pattern`: glob (`*`, `?`); defaults to `*` (everything).
- `channel`: `discord | ntfy | all`.
- Loaded into memory by `notify_scan()` and applied **after** the existing alert-key derivation. No change to detection logic — only routing.
- New `GET/POST /api/notify/rules` for the UI.

- A *Routing rules* panel inside the existing **Alerts** tab.
- Table view: kind · pattern · channel · min severity · enabled · 🧪 test.
- Inline add / edit / remove.
- "Test this rule" sends a sample alert that *would* match it (verifies wiring end-to-end).

## Acceptance
- [ ] Rule `container ollama-* → ntfy @ critical`: an unhealthy `ollama-worker` fires on ntfy only; an unhealthy `postgres-prod` still fires on both via the default rule.
- [ ] No rules configured → behaviour identical to v0.7.0.
- [ ] Rules survive container rebuild (`./data/gpu.db` bind mount).

## Pointers
- Existing notifier: search `app.py` for `_NOTIFIED`, `notify_scan`, `_post_json`. Dispatch already groups by channel.
- Settings persistence pattern is right there too (`settings` table, `get_settings`, `save_settings`).

🟡 medium · 🛠️ help wanted · 🔔 alerts

Mr-Neutr0n:
Fixes #24

Added per-service notification routing rules — new `notification_rules` DB table, `_apply_rules()`/`_dispatch_to_channels()`/`_match_kind()`/`_alert_name()` helpers, modified `_emit()` and `notify_scan()` to load and apply rules (with OOM handler also covered), `GET/POST /api/notify/rules` and `POST /api/notify/rules/test` API endpoints, and an inline-editable routing rules table UI in the Alerts tab with add/edit/delete/test per-row. No rules → identical v0.7.0 behavior.

Local test infra unavailable in CI sandbox.

---
This change was prepared with AI assistance under human direction and review.


SikamikanikoBG:
Thanks for tackling #24 — per-service routing is a genuinely useful control, and the core is sound: `_apply_rules()` returning `None` to preserve the legacy "all channels" behaviour is exactly right, the SQL is parameterized, and the `LOCK` usage matches the codebase.

A few things need attention before it can land on `next` (which has moved a fair bit — uptime monitoring, the cost heatmap and the GPU work all shipped since this PR). None of these are hard — mostly rebase mechanics plus styling/i18n alignment:

**Must-fix**
1. **Rebase onto `next`.** Small/mechanical — about 3 conflict sites: the `import` line (keep `fnmatch`), the two new routes (keep both alongside the new self-update/uptime routes), and the alerts-pane HTML hunk (the heading is now `<h2 data-i18n=…>` and the pane was restructured).
2. **Route the new uptime alerts too.** `notify_uptime()` now fires DOWN/recovery/slow alerts via `_emit(...)`, but those calls don't pass `rules=rules`, and `_match_kind()` has no `uptime` case — so uptime alerts currently bypass your routing entirely. Please thread `rules=rules` through those `_emit` calls and add an `uptime` kind (or fold it under `host`).
3. **`var(--b)` → `var(--bd)`.** `--b` isn't defined anywhere, so the container/table/thead borders and the JS `tr.style.borderBottom` render with the wrong colour in both themes. The dashboard's border token is `--bd`.
4. **Use a real heading.** The section title is a `<div>`, so `applyIcons()` doesn't convert the 📋 to an SVG and it renders as a raw emoji. Make it `<h2>📋 Routing rules</h2>` (or use `sic()`), matching every other card heading.
5. **Add i18n.** The new UI is 100% hardcoded English while the rest of the Alerts pane uses `data-i18n` / `I18N.t(...)`. Please localize the labels, captions, buttons and the JS `confirm()` string.

**Should-fix**
- Replace the inline-style soup with the existing classes (`.kpi` / `.chip` / `.badge` / `.cap`), and add `aria-label`/`title` to the icon-only row buttons.

Happy to re-review once it's rebased and the styling/i18n is brought in line. Thanks again!

SikamikanikoBG:
Merged — thank you. This is exactly what #24 needed.

Re-reviewed the rebased branch and the important stuff all landed: the uptime alerts now thread `rules=rules` and `_match_kind()` has a proper `uptime` case, so the routing no longer bypasses uptime (and there's a test proving it). The heading is a real `<h2>` so `applyIcons()` picks up the 📋, the borders use `--bd`, and `_apply_rules()` returning `None` keeps the legacy "all channels" default intact.

The only thing left was i18n: the markup was already wired with `data-i18n`/`I18N.t`, but the `rules.*` keys were never added to the locale files, so a zh-CN user saw the whole section in English. I pushed a small maintainer follow-up adding all 27 keys to `en.json` and `zh-CN.json` (full parity) — your feature logic is untouched.

Appreciate you sticking with the rebase and the styling pass. Nice work.

PR #156, opened 2026-06-16
fix: preserve legend filter state across auto-refresh for network chart (#155)

### LINKED ISSUE #155: Network › Throughput: isolating one NIC is tedious and the legend filter resets on every auto-refresh
### Summary

Reported by @itcat672 in #148. 🙏 On a host with ~6 network interfaces, the **Network → Throughput** chart plots every NIC as two series (↓ in / ↑ out), so ~12 lines at once. Two pain points:

**1. Isolating a single NIC is tedious (enhancement).**
To view just one interface you currently have to click off the in+out legend entries of the other 5 NICs — ~10 clicks.

**2. The legend filter resets every ~15s (bug).**
Even after manually hiding the others, the global auto-refresh rebuilds the chart and wipes the legend selection, so the isolated view only lasts a few seconds before everything reappears.

### Root cause (confirmed)

`static/dashboard.html`:

- Auto-refresh (every 15s) re-renders the tab:
  ```js
  setInterval(()=>{
    if(!document.getElementById('auto').checked) return;
    ...
    if(TAB==='network') renderNetwork();   // -> renderNetIO()
  },15000);
  ```
- `renderNetIO()` calls `mk('netchart', …)`, and `mk()` does `charts[id].destroy()` then `new Chart(...)`. Destroying + recreating the chart resets each dataset's `hidden` flag to its default, so Chart.js's built-in legend toggle state is lost on every refresh.

### Suggested direction (open to discussion)

- **Fix the reset:** preserve per-series visibility across refresh — capture the current `hidden` state (keyed by series label / `iface`) before `destroy()` and re-apply it after rebuild; or update the existing chart's `datasets[].data` in place + `chart.update()` instead of destroy-and-recreate.
- **Easier isolation:** add a "solo" affordance — e.g. a per-NIC selector/dropdown, or a custom legend `onClick` where a normal click toggles and a modifier/long-press isolates one NIC (hide all others). Optionally fold ↓/↑ of one NIC into a single legend toggle.

### Good first issue?

Reasonably — it's contained to `renderNetIO()` + `mk()` and the auto-refresh block. Requires a little Chart.js (v4.4.1) familiarity.

laishettikarthik-tech:
Fixes #155

## Problem
On hosts with multiple NICs, the Network → Throughput chart plots ~12 series
at once. When a user manually hides series via the legend to isolate one NIC,
the auto-refresh (every 15s) calls mk() which does destroy() + new Chart(),
wiping all legend toggle state — so the isolated view resets every 15 seconds.

## Root cause
`mk()` had no memory of which datasets were hidden before destroy():
```javascript
function mk(id,cfg){ if(charts[id])charts[id].destroy(); charts[id]=new Chart(...,cfg); }
```

## Fix
Capture each dataset's hidden state (keyed by label) before destroy(), then
re-apply it after the new chart is created:

- Before destroy: record `hidden[ds.label] = !chart.isDatasetVisible(i)`
- After new Chart(): find each label and call `chart.hide(idx)` if it was hidden

## Files changed
- `static/dashboard.html` — updated `mk()` function only

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # 🛰️ HomeLab Monitor
 
 [![GitHub stars](https://img.shields.io/github/stars/SikamikanikoBG/homelab-monitor?style=social)](https://github.com/SikamikanikoBG/homelab-monitor/stargazers)
+[![Discord](https://img.shields.io/badge/Discord-join%20the%20chat-5865F2?logo=discord&logoColor=white)](https://discord.gg/yFEf3N2mb9)
 [![version](https://img.shields.io/github/v/release/SikamikanikoBG/homelab-monitor?color=blue&label=version)](CHANGELOG.md)
 ![license](https://img.shields.io/badge/license-MIT-green)
 ![docker](https://img.shields.io/badge/deploy-docker--compose-2496ED?logo=docker&logoColor=white)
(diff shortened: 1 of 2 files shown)


SikamikanikoBG:
Merged into `next` — thank you, @laishettikarthik-tech! 🎉

Nice fix. You nailed the root cause (the 15s auto-refresh calling `mk()` → `destroy()` + `new Chart()`, which wipes the legend toggle state) and the snapshot-by-label / re-apply approach is the right shape — it's generic, so every chart keeps its hidden-series state across refreshes, not just the network one. One small thing on my side: the PR was opened against `main`, so I retargeted it to `next` (our integration branch) before merging — nothing for you to do. It'll ship in the next release.

If you'd like to hang out with other folks building and using HomeLab Monitor — share ideas, grab issues, or just say hi — you're very welcome in the Discord: https://discord.gg/yFEf3N2mb9

Thanks again for squashing this one. 🙏

PR #190, opened 2026-06-22
feat: add public read-only status page

siva23367:
## Checklist

- [x] **Base branch is `next`** (the integration branch), not `main`. Change the
- [x] Branched off the **latest `next`** (`git pull origin next` before
- [x] Tested locally with `docker compose up -d --build` and the dashboard
      behaves as expected.
- [x] No version bump in this PR — version bumps are handled by the maintainer
      at release time.
- [x] For a new monitor / probe: followed the **"add a monitor" pattern** in


SikamikanikoBG:
Thanks for this — a public status page is a nice idea and clearly a lot of work went into it. I can't merge it as-is yet, primarily for a security reason that's important to flag clearly:

**Critical — data exposure**
The `/public` route is **unauthenticated** and the page fetches the full internal `/api/health`, which includes host filesystem paths (`HOST_ROOT`, `DB_PATH`, the docker socket path), the **process list**, the full **container and systemd-service inventory** (with names), **pending OS updates**, hostname and version. This tool runs with broad host access and is meant to live behind a LAN/VPN — so a public endpoint must expose *only* coarse status, never any of that.

**Must-fix**
1. **Add a dedicated sanitised endpoint** (e.g. `/api/public-status`) that returns only roll-ups: an overall `ok/warn/crit` plus bare counts (e.g. "containers 12/12 up", aggregate CPU/RAM/GPU %). No paths, no process list, no host/container/service names, no IPs/ports, no version, no `os_updates`.
2. **Make it opt-in and off-by-default** — gate both the route and the page behind an env flag (e.g. `PUBLIC_STATUS=1`) and return 404 when unset.
3. **Fix the data contract.** The page currently reads fields that don't exist in `/api/health` (`overview.cpu_pct`, `containers.items`, `hosts`, …) — the real shape is a list-of-cards `overview`, `now.gpu`, `docker.containers`, `systemd.services` — so it renders mostly blank against the live API.
4. **Match the dashboard's design language.** Right now it's a separate look: its own hardcoded colour tokens (dark-only, no `[data-theme]` light support), emoji instead of the stroke-SVG icon system (`svgIc()`/`applyIcons()`), a different font, and duplicated `.card`/`.chip`. Please reuse the existing CSS custom properties, support both themes, and use the SVG icon family so it feels like the same product.
5. **Add a test** asserting the public endpoint omits the sensitive keys, and rebase onto `next` (the `api_health()` function has changed).

The concept is welcome — it just needs the sanitised endpoint + opt-in gate + a restyle before it's safe to ship. Thank you for the effort, and happy to help shape the safe version.

PR #191, opened 2026-06-22
added email , slack and generic webhooks

Jishnu-Prasad888:
## What this changes

Adds email/Slack/generic webhook alert polish: validates email settings, improves SMTP sending (proper headers, STARTTLS/implicit TLS), reinstates docker notifier binding to prevent crashes, and documents TLS expectations. Expands notification tests with validation and docker regression coverage.

## Checklist

- [ ] **Base branch is `next`** (the integration branch), not `main`. Change the
- [ ] Branched off the **latest `next`** (`git pull origin next` before
      behaves as expected.

## Notes for the reviewer

- Unit tests added for email validation/notifier and docker notifier regression; `python3 -m unittest tests.test_notify` currently fails here due to missing dependencies (Flask/pytest). Please install project test deps or run in the standard dev container verify.


SikamikanikoBG:
Thanks for this — it's a clean, well-structured feature and genuinely close to mergeable. The new channels reuse the existing `_post_json` / `dispatch_alert` patterns faithfully, secrets are added to `SETTING_SECRETS` (so they're masked in `_public_settings`), the URL keys get the same scheme/host validation as the existing webhooks, and the `tests/test_notify.py` coverage (STARTTLS vs implicit-TLS vs none, auth on/off, masking, port validation, the docker-block regression) is solid. The settings UI also matches the dashboard's field/`.flabel`/`.cap` pattern natively.

Two small things before it can land on `next`:

**Must-fix**
1. **Rebase / resolve one conflict.** Against current `next` there's a single one-line conflict — the top-level `import` (next added `urllib.error`; this PR adds `smtplib` + `email.message`). Just union them onto one line. Everything else (dashboard.html, locales, tests, docs) auto-merges.
2. **Wire up the i18n keys.** The 8 new keys in `en.json` / `zh-CN.json` (`set.email_host`, `set.slack_url`, `set.webhook_url`, …) aren't referenced anywhere — the new field labels are hardcoded English, so a zh-CN user sees English for the email/Slack/webhook rows while the rest of the Alerts panel localizes. Add `data-i18n="…"` to the new `.flabel` divs (and ideally the captions), or drop the unused keys.

**Nice-to-have**
- The PR description talks about reinstating a docker notifier binding, which doesn't match the diff (a clean additive feature) — worth tidying.

Once the import is resolved and the labels are wired, this is good to go. Thank you!

Jishnu-Prasad888:
Implemented the mentioned changes : 
- resolve the top-level import conflict so both urllib.error and the new SMTP/email modules ship together
- hook the email, Slack, and generic webhook alert fields up to their locale keys for proper translations
- refresh the changelog entry to describe the new alert delivery routes instead of the old docker notifier note

SikamikanikoBG:
Merged — thank you, this is a clean addition. Email/Slack/generic-webhook is a genuinely useful trio and it reuses `_post_json`/`dispatch_alert` faithfully, the secrets are masked in `SETTING_SECRETS`, the URL keys get the same scheme/host validation as the existing webhooks, and the `test_notify.py` coverage is thorough (STARTTLS/implicit/none, auth on/off, masking, the docker-block regression). Both must-fixes confirmed done.

A few things I sorted out before merging. The branch was cut before 0.18.0–0.21.0 shipped, so its `CHANGELOG.md` had dropped those four released entries and replaced the top with `[Unreleased]`. I restored the full history from `next` and kept your alert-delivery note under a fresh `[Unreleased]` section (additive only). While I was in there I keyed the last three hardcoded strings (the Email (SMTP) header, the TLS label, the webhook caption) so the whole panel localizes.

There was also a merge seam with #151 (per-service routing rules), which landed just before this: the new per-service rules only knew about Discord/ntfy/Telegram, so a rule would have silently skipped your email/Slack/webhook channels. I wired the three new channels into the rule dispatcher and the "all" expansion, added them to the rule channel dropdown, and added a test so a rule can now route to any of them. Your channel code is unchanged — this was just connecting the two features.

Thanks for the solid tests and the quick turnaround on the earlier review.

PR #212, opened 2026-06-27
feat(uptime): maintenance windows — silence alerts during planned work (#167)

### LINKED ISSUE #167: Checks: maintenance windows — silence alerts during planned work
## What
Schedule a **maintenance window** (one-off or recurring) covering selected checks. While it's active, those checks don't fire down/recovery alerts and show as **in maintenance** (not down) on the dashboard and the status page.

## Why
Planned reboots, upgrades and backups shouldn't page you. Without this, the first thing people do after wiring up alerts is mute everything by hand during maintenance — or, worse, learn to ignore alerts. Small, and it's what makes the alerting trustworthy enough to leave on.

### Backend (`app.py`)
- A `maintenance` table (checks, start, end, optional recurrence, note); suppress alerts while a window is active.

- A simple "schedule maintenance" form and an **in maintenance** badge (blue) on affected checks.

## Acceptance
- [ ] A check inside an active window shows **in maintenance**, not down, and sends no alerts.
- [ ] Alerts resume automatically when the window ends.

> Stretch item of **0.18 — Up & responding**. Pairs with down/recovery alerts.

1HazyOne707:
## Summary

Closes #167

Planned reboots and backups shouldn't page you. This adds maintenance windows that silence alerts for selected checks/services while active, showing them as **in maintenance** (blue) instead of down on the dashboard.

## How it works

- A `maintenance_windows` table stores windows with a `kind` (container, systemd, uptime, disk, gpu, `*`) and fnmatch `pattern` so you can target everything (`*`) or a specific check by name/id
- `_in_maintenance(kind, name)` is called in `_emit()` before the edge-trigger fires — if the key matches an active window, the alert is dropped silently
- `notify_uptime()` checks maintenance per-check before firing down/recovery/cert/slow alerts; the check is skipped entirely while in maintenance
- Supports one-off, daily, and weekly recurrence

### Backend (`app.py`)
- `maintenance_windows` table with schema migration
- `_in_maintenance(kind, name)` — fnmatch matching, recurrence-aware
- Suppression injected into `_emit()` and `notify_uptime()`
- `in_maintenance` flag on every check in `uptime_overview()`
- `GET/POST /api/maintenance` and `DELETE /api/maintenance/<id>`

- `.badge.maint` CSS (blue, consistent with existing badge system)
- `upCard()` shows blue **In maintenance** pill when check is covered

### Tests (`tests/test_uptime.py`)
- 11 new tests: create/list/delete, active/expired, wildcard kind+pattern, kind mismatch, daily recurrence, `_emit()` suppression, emit fires outside maintenance, `uptime_overview` flag
- All 68 tests in `test_uptime.py` pass


SikamikanikoBG:
Thanks for tackling maintenance windows — the schema/migration and matching approach are solid, and the auth model is consistent with the rest of the app. Found a few things I'd like fixed before merging:

1. `POST /api/maintenance`: `int(end_ts)` / `int(start_ts)` aren't guarded, so a non-numeric value throws an unhandled `ValueError` → a raw 500 instead of a clean 400. The rest of the app (see `_validate_uptime_check`) wraps these conversions in try/except; can you do the same here?
2. In `notify_uptime()`, the cert-alert clear on the in-maintenance branch uses `cert_key if "cert_key" in dir() else ...` — but `cert_key` is a leftover local from a *previous* iteration's `elif state == "up"` branch, so once any check earlier in the loop hits that branch, this clears an unrelated check's `uptime:cert:<id>` key instead of the current one's. Simplest fix: just `_clear(f"uptime:cert:{cid}")` directly, no `dir()` check needed.
3. Neither `_in_maintenance` nor the create endpoint caps window length for daily/weekly recurrence. A daily window accidentally spanning >24h (or a weekly one >7d) becomes permanently active — with `kind=*`/`pattern=*` that's a silent, indefinite suppression of every alert app-wide. Worth a validation check (`end_ts - start_ts <= 86400` for daily, `<= 604800` for weekly) so a misconfig fails loud instead of silent.

Happy to take another look once these are in — the 11 new tests are a good start, they just don't yet exercise the HTTP route or the weekly/oversized-window paths, which is how these slipped through CI.

1HazyOne707:
Pushed fixes for all three points:

1. `start_ts`/`end_ts` now wrapped in try/except — non-numeric values return a clean 400 instead of an unhandled ValueError.
2. Cert-key clearing in `notify_uptime()` now calls `_clear(f"uptime:cert:{cid}")" directly, no more leftover `dir()` check — added a regression test confirming it clears the correct check's key, not a stale one from an earlier loop iteration.
3. Added window-length caps: daily recurrence can't exceed 86400s, weekly can't exceed 604800s — both return a 400 instead of silently going indefinite.

72 existing tests + 3 new ones (numeric guard, oversized window rejection, cert-key correctness) all passing.

SikamikanikoBG:
Merged. 🎉

This one went in clean — 337/338 tests green (the single miss is the known Windows-only `test_backup.py` flake, nothing to do with your change), zero conflicts against `next`, and the code itself is tidy: solid input validation on the new `/api/maintenance` routes and properly parameterized SQL throughout. No notes.

And it's a genuinely useful feature. Silencing alerts during planned work was an obvious gap — every time I did maintenance I'd get a wall of noise and half-wonder if I'd broken something real. This closes that properly.

Fourth one now, and the quality's been consistent every time. Thanks for keeping at it — really appreciate having you on this. 🙏

1HazyOne707:
Thanks, Arsen — really glad it landed clean, and I appreciate you saying that. Genuinely enjoyed working on this, and I'm grateful for the trust and the honest reviews along the way.

Now that this is merged into next, I'll rebase #217 (status page) on top so its diff only carries the status-page delta, per your earlier note.

Separately from the code — I wanted to ask directly: are you aware of any paid contribution work, or even a role, that might be a fit? I'm in a tough financial stretch right now and would love to find a way to do this kind of work full-time. No pressure at all — just figured I'd ask given how well this has been going.

PR #217, opened 2026-06-27
feat: status page — shareable read-only 'is everything up?' page (#166)

### LINKED ISSUE #166: Status page — a shareable, read-only "is everything up?" page
## What
A public, read-only **status page** built from the checks: each one (optionally grouped into sections like "AI" or "Core services") with its current state, uptime % and heartbeat bar. Custom title, the new radar logo, served on its own route — and no host internals leaked (same spirit as Share: nothing private).

## Why
It's the one outward-facing artefact people screenshot and share — and a clean way to tell housemates or teammates "everything's fine" without handing them the whole dashboard. It's also free marketing: a tasteful status page carrying our radar mark travels well.

### Backend (`app.py`)
- `GET /status` (+ `GET /api/status`) returning only safe, check-level data — no host names/IPs unless explicitly opted in.
- Settings: page title, which checks are public, optional grouping.

- A standalone, theme-aware status page (reuse the heartbeat bar + uptime chips); link it from Settings.

## Acceptance
- [ ] `/status` renders current state + uptime % for the chosen checks, no auth needed.
- [ ] Only opted-in checks appear; no host internals leak.
- [ ] Works in light & dark, down to 360 px.

> Stretch item of **0.18 — Up & responding**. Builds on the Checks engine + uptime history.

1HazyOne707:
## Summary

Closes #166

A public, read-only status page built from the uptime checks — no auth needed, no host internals leaked. Pairs naturally with the maintenance windows from #212 (merged/pending): checks under an active window show as **In maintenance** (blue) rather than down.

## How it works

- `GET /status` serves a standalone `status.html` (404 when disabled)
- `GET /api/status` returns only safe, check-level data — label, state, uptime %, latency, heartbeat strip, cert days, in_maintenance flag. No targets, no host names, no IPs
- `overall` field: `operational` / `degraded` / `maintenance` / `unknown`
- Three settings: `status_page_enabled`, `status_page_title`, `status_page_checks` (`*` or comma-separated check IDs)

### Backend (`app.py`)
- Three new settings with defaults
- `GET /status` — serves `status.html`, 404 when disabled
- `GET /api/status` — public JSON, safe check-level data only

### Frontend (`static/status.html`)
- Standalone page, no dashboard dependency
- Theme-aware (dark/light via `prefers-color-scheme`)
- Heartbeat bar, uptime %, latency, cert pill per check
- Blue **In maintenance** pill when covered by a maintenance window
- Overall banner: green / red / blue / grey
- Auto-refreshes every 30s
- Works down to 360px, no external assets

### Tests (`tests/test_uptime.py`)
- 4 new tests: disabled=404, enabled=200+title, degraded when check down, check ID filtering
- 120 tests passing across `test_uptime.py` + `test_notify.py`


SikamikanikoBG:
Thanks for this — the status page itself (private-data handling, XSS escaping, auth model) all looks right. Holding off on merging though: this diff carries the entire maintenance-windows feature from #212 along with it (the `maintenance_windows` table, `_in_maintenance`, the `/api/maintenance` routes, the `.badge.maint` CSS — all of it), which suggests this branch was built on top of #212 rather than off current `next`. `next` doesn't have any of that yet, so merging #217 as-is would land #212's code as a side effect, and merging #212 afterward would then collide on duplicate route registrations (Flask throws on the second `/api/maintenance` registration).

Once #212 lands (after the fixes I left there), could you rebase this onto the updated `next` so the diff only carries the status-page delta? Separately, a small bug worth fixing in `static/status.html`'s `load()`: `!j.ok===false` parses as `(!j.ok)===false`, which is always false exactly when `j.ok` is false — so the "show the real server error" branch never fires and users just see the generic empty state instead of the actual error. Should be something like `!j.ok && j.error`.

SikamikanikoBG:
First off — thank you for this. The work is solid and I can see the care that went into it. This isn't a quality call at all, so please read the rest in that spirit: it's genuinely just unlucky timing.

Here's what happened. #217 branched off #212 and carries a full standalone public status page — the `/status` route, `/api/status` endpoint, `static/status.html`, gated behind the new `status_page_enabled` / `status_page_title` / `status_page_checks` settings. The thing is, since you opened this on 2026-06-27, v0.22.0 shipped a public status page from another contributor (@siva23367): `/public` + `/public/<id>`, `/api/public-status` + `/api/public-status/<id>`, gated by the `PUBLIC_STATUS` env var. That one's actually more built out — per-service detail pages, a 14-day incident history timeline, uptime across 24h/7d/30d/90d, and TLS cert expiry.

So I don't want to merge #217 as-is, and here's the honest reasoning: shipping two parallel public status pages — different URLs, different settings, same job — would confuse users and, more importantly, it doubles the privacy-sensitive data-filtering surface we'd have to keep correct on both. One is already something I want to watch like a hawk; two is asking for a leak.

**But** — your version has one thing the existing page doesn't, and it's a good idea: the **"In maintenance" blue pill** when a check is covered by a maintenance window from #212. That integration is exactly the kind of thing the `/public` page should have. I also prefer your **config-via-Settings-checkbox** approach over the env-var-restart the current page needs.

So the ask, if you're up for it: instead of a competing page, would you consider a follow-up PR that brings the **"In maintenance" badge onto the existing `/public` page**, and (separately, if you want) a Settings toggle so it doesn't need an env-var restart? That'd land the genuinely novel part of this work where it fits.

I'll leave #217 open — closing it is your call (or Arsen's), not something I want to do over your head. Just wanted to explain plainly why it's not going in right now rather than let it sit.

Thanks again — you've been one of the most reliable contributors on this project and I'd hate for this to read as a brush-off. Keep them coming. 🙏

1HazyOne707:
HI @SikamikanikoBG Thanks for the detailed feedback and for taking the time to explain the reasoning here — it’s appreciated.
I understand why #217 doesn’t fit with the current /public implementation and I’ll leave it as-is.
Just confirming: do you still want me to handle the follow-up PR for the maintenance badge + settings toggle, or is that already covered? I’m happy to work on it if it’s still needed.

SikamikanikoBG:
@1HazyOne707 Closing the loop here, with apologies for the slow reply: yes, the follow-up we discussed is done — you did it. #231 (the maintenance-window badge on `/public`) merged, and a Settings toggle exists now too (`public_status_enabled`, alongside the `PUBLIC_STATUS` env var), so both improvements from this PR live where they belong. With that there's nothing left here to land, so I'm closing this one as superseded — the work wasn't wasted, it directly shaped what shipped. Thanks for taking the timing collision so gracefully, and for #236 and everything since.

PR #222, opened 2026-07-01
feat: icons for the sidebar nav

SikamikanikoBG:
The left nav was plain text, unlike every content page which already carries an SVG icon in its own h2 header. Added one icon per nav row, reusing the exact SVG each tab already draws in its header (same `ICONS` registry, keyed by the emoji that used to prefix it) so the row visually matches what you land on. The four tabs with no existing header icon (Overview, Disks, Disk I/O, Settings) get a small dedicated glyph in the same stroke family: gauge, hard-drive, activity pulse, and gear.

No backend changes. Full suite: 286 passed, 2 pre-existing failures unrelated to this change (confirmed pre-existing on unpatched `next`).


SikamikanikoBG:
Verified live: pulled this branch and screenshotted the sidebar in both light and dark themes at desktop width — icons render cleanly, selected-state accent color and hover states all work as expected. No visual regressions.

PR #230, opened 2026-07-07
refactor: extract app.py monolith into backend/ module tree (Phases 1–4)

pehota:
## Summary

Executes the accepted [RFC](design/rfc-refactor.md) — breaks the ~7600-line monolithic `app.py` into a structured `backend/` module tree. All 54 routes, all collectors, probes, notify logic, and DB access are now in separate modules. `app.py` is a thin wiring layer.

- **Rollup tables** (`samples_1h`, `net_samples_1h`) reduce long-range query cost; cost/heatmap endpoints already migrated to them
- **Raw retention** dropped 180d → 48h (rollups retain full history) — ⚠️ irreversible on first boot
- **Alert edge-state** (`_NOTIFIED`, `_uptime_down_since`) persisted to DB — survives restarts, no duplicate/missed alerts
- **Worker isolation** — each worker in its own daemon thread; watchdog logs stalls
- **No silent swallows** — all broad `except` blocks in `backend/` now log; AST-based CI gate prevents regression

## Phases

| Phase | Description |
|-------|-------------|
| 0.1–0.2 | Snapshot tests for all 54 endpoints; CI gate |
| 1.1 | `PRAGMA busy_timeout=5000` |
| 1.2a | Rollup tables + backfill + collector upsert |
| 1.3 | Cost/heatmap readers migrated to `samples_1h` |
| 1.2b | Raw retention 180d → 48h ⚠️ |
| 2.1–2.3 | `backend/db/`: thread-local conn factory, migration runner, repo shims |
| 3.1–3.4 | `backend/probes/`, `backend/collectors/`, `backend/notify/`, `backend/api/` blueprints |
| 4.1 | All `.execute()` calls migrated to repo functions |
| 4.2 | Alert edge-state persisted across restarts |
| 4.3 | Per-worker heartbeat + watchdog thread |
| 4.4 | Logged exception handling; CI gate via AST |

## Test plan

- [ ] 452 tests pass (`pytest tests/ -q`)
- [ ] Snapshot suite byte-identical (71 snapshots)
- [ ] `test_no_silent_swallow` passes (AST gate)
- [ ] `test_worker_isolation` passes (stall isolation proven)
- [ ] First boot: verify `notified_keys` + `uptime_down_since` tables created
- [ ] Restart test: arm an alert, restart, confirm no duplicate fires
- [ ] ⚠️ Raw samples older than 48h will be trimmed on first boot — ensure rollup backfill ran before upgrading prod


SikamikanikoBG (changes requested):
Huge effort here, and it shows — the phased approach (snapshot gate → rollups → db/probes/collectors/notify/api extraction → edge-state persistence → worker isolation → silent-swallow gate) is genuinely well structured, and 448 tests + 71 byte-identical snapshots is a serious safety net for a refactor this size.

I pushed 3 small fixes straight to the branch to get CI further:
- Dockerfile wasn't copying the new `backend/` package into the image — container crashed on boot with `ModuleNotFoundError: No module named 'backend'`.
- The new `snapshot` CI job only installed `flask`, so it died before a single test ran (`No module named pytest`), and was also missing `prometheus_client` so `/metrics` didn't match its own baseline.
- `backend/db/__init__.py` did `import app as _app` at module scope just to read `DB_PATH` — fixed by reading the env var directly (same default `app.py` uses).

That last one is the thread that led somewhere bigger, and this one I didn't fix because it's an architecture call, not a one-liner: the lazy `import app as _app` pattern is used 91 times across `backend/api`, `backend/collectors`, `backend/probes`, `backend/notify`. It's safe under pytest (which imports `app` as a real module), but `launch.py` boots the dashboard as `python /app/app.py` — that registers the running code as `__main__`, not as `sys.modules['app']`. So the first `import app as _app` anywhere (including inside the `/healthz` handler itself) re-executes `app.py` from scratch as a second module object, re-registers the Prometheus gauges at module scope, and blows up:

```
ValueError: Duplicated timeseries in CollectorRegistry: {'homelab_gpu_vram_used_mb'}
uptime_worker error: module 'app' has no attribute '_uptime_tick'
host_poller error: module 'app' has no attribute 'list_hosts'
```

Net effect: the container can't pass its own `HEALTHCHECK` on a clean boot. Worth sorting before this goes into `next` — probably having `launch.py` `import app` and call into it directly instead of subprocessing `python app.py`, so `app` only ever loads once under its real name. Happy to pair on it.

Separately, 2 snapshot tests (`api_cost`, `api_cost_heatmap`) fail on CI — the cost/heatmap bucketing uses `time.localtime()`/`time.mktime()`, so it's tied to the host's local timezone. The checked-in baselines look like they were generated on a non-UTC machine, so they mismatch on GitHub's UTC runners. Probably wants `TZ=UTC` pinned in the test setup (or wherever these get regenerated) plus a rebaseline.

Really nice work overall — just holding off on the merge until the container boots clean.

pehota:
> Huge effort here, and it shows — the phased approach (snapshot gate → rollups → db/probes/collectors/notify/api extraction → edge-state persistence → worker isolation → silent-swallow gate) is genuinely well structured, and 448 tests + 71 byte-identical snapshots is a serious safety net for a refactor this size.
> 
> I pushed 3 small fixes straight to the branch to get CI further:
> 
> * Dockerfile wasn't copying the new `backend/` package into the image — container crashed on boot with `ModuleNotFoundError: No module named 'backend'`.
> * The new `snapshot` CI job only installed `flask`, so it died before a single test ran (`No module named pytest`), and was also missing `prometheus_client` so `/metrics` didn't match its own baseline.
> * `backend/db/__init__.py` did `import app as _app` at module scope just to read `DB_PATH` — fixed by reading the env var directly (same default `app.py` uses).
> 
> That last one is the thread that led somewhere bigger, and this one I didn't fix because it's an architecture call, not a one-liner: the lazy `import app as _app` pattern is used 91 times across `backend/api`, `backend/collectors`, `backend/probes`, `backend/notify`. It's safe under pytest (which imports `app` as a real module), but `launch.py` boots the dashboard as `python /app/app.py` — that registers the running code as `__main__`, not as `sys.modules['app']`. So the first `import app as _app` anywhere (including inside the `/healthz` handler itself) re-executes `app.py` from scratch as a second module object, re-registers the Prometheus gauges at module scope, and blows up:
> 
> ```
> ValueError: Duplicated timeseries in CollectorRegistry: {'homelab_gpu_vram_used_mb'}
> uptime_worker error: module 'app' has no attribute '_uptime_tick'
> host_poller error: module 'app' has no attribute 'list_hosts'
> ```
> 
> Net effect: the container can't pass its own `HEALTHCHECK` on a clean boot. Worth sorting before this goes into `next` — probably having `launch.py` `import app` and call into it directly instead of subprocessing `python app.py`, so `app` only ever loads once under its real name. Happy to pair on it.
> 
> Separately, 2 snapshot tests (`api_cost`, `api_cost_heatmap`) fail on CI — the cost/heatmap bucketing uses `time.localtime()`/`time.mktime()`, so it's tied to the host's local timezone. The checked-in baselines look like they were generated on a non-UTC machine, so they mismatch on GitHub's UTC runners. Probably wants `TZ=UTC` pinned in the test setup (or wherever these get regenerated) plus a rebaseline.
> 
> Really nice work overall — just holding off on the merge until the container boots clean.

Addressed all issues and fixed the pipeline.
Can you please review again?

SikamikanikoBG:
Huge work here. Splitting a 7600-line monolith into `backend/` modules, adding the rollup tables, persisting alert state across restarts, and the snapshot-test CI gate — that's a lot of careful, disciplined engineering. Thanks for pushing it through all the CI churn today. Merged, and really appreciate it.

PR #231, opened 2026-07-08
feat: show maintenance-window badge on public status page

1HazyOne707:
Follow-up to the discussion on #217 — brings the 'In maintenance' badge onto the existing /public status page instead of shipping a competing one.

## Changes
- `_public_monitor()` now computes `in_maintenance` via the existing `_in_maintenance()` helper (same pattern used by the private dashboard's `uptime_overview()`)
- `_public_overall_status()` returns a new `maintenance` state when nothing is down but something is under an active window
- Frontend already had `.maint` CSS and `stCls`/`stLabel` support scaffolded — this wires real data through to it
- New test: `test_public_monitor_shows_in_maintenance`

Base is `next` directly (not built on #212 or #217) to avoid the route-collision issue flagged there.


SikamikanikoBG:
Thanks for this — clean, small change with good test coverage. One thing: it branched before #230 (the backend/ refactor) merged into `next`, which added a mandatory snapshot-test CI gate — so CI here never exercised that gate. I test-merged locally against current `next`: no conflicts, and the existing snapshot baselines hold. Could you rebase onto current `next` so CI picks up the new snapshot job? Happy to merge right after.

1HazyOne707:
Rebased onto next and pushed — turned out the public-status routes had moved to backend/api/uptime_api.py in the meantime, so I updated the toggle to match that, and rebaselined two settings snapshots that legitimately changed shape with the new field. All checks green now.

SikamikanikoBG:
Thanks @1HazyOne707 — merged! The maintenance-window badge is a small touch that makes a real difference. When something's down on purpose, readers instantly know it's planned. Great addition, appreciate you sending it in. (Heads up: I merged latest `next` into your branch to clear a small settings-snapshot conflict with the retention PR that landed just before — nothing you needed to do.)

1HazyOne707:
Following up on this — after it merged I noticed the maintenance flag only made it into `_public_monitor()` and `uptime_overview()`, not into three related spots:

1. `_public_status_detail()` (the per-service detail endpoint) doesn't return `in_maintenance` at all.
2. `_public_monitors_summary()` doesn't include a maintenance count alongside up/down.
3. `_public_overall_status()` still reports `crit` for a check that's down but fully covered by an active maintenance window, rather than `maintenance`.

Was that intentional — e.g. keeping the detail page and overall-status logic simpler on purpose — or just not the direction you'd want? If it's a genuine gap, I've got a fix + new test coverage ready (also caught a `_wipe()` bug in the test fixture that let a maintenance window leak across tests). Happy to open a small follow-up PR if useful.

PR #233, opened 2026-07-10
feat: configurable data retention + misc hardening fixes

pehota:
## Summary

- Add **General** settings tab with a `retention_days` field (1–3650 days); value is read live each pruning cycle so changes apply within ~6 minutes without a restart.
- Fix Prometheus gauge double-registration by caching gauges in a module-level `_GAUGES` dict (prevents `ValueError` on Flask debug-reloader / double-import).
- Fix docker policy refresh using the wrong TTL constant (`_DOCKER_ENRICH_TTL` → `_DOCKER_POLICY_TTL`).
- Improve container self-ID detection via `/proc/self/cgroup` (cgroups v1 full 64-char ID); falls back to `HOSTNAME` env var for v2 / non-container environments.
- Guard `POST /api/hosts/<name>/run` behind `ENABLE_CONTROLS` check (returns 403 when controls are disabled).
- Add a confirm dialog before changing a container's restart policy.
- Read retention window outside `LOCK` in `sample_once` to prevent deadlock (`get_settings` is also non-reentrant on `LOCK`).

## Test plan

- [ ] Open Settings → General, change retention days, verify value persists after save
- [ ] Confirm pruning respects the new value after ~6 minutes
- [ ] Restart Flask with `--debug` — verify no `ValueError` from Prometheus on reload
- [ ] With `ENABLE_CONTROLS=0`, confirm `POST /api/hosts/.../run` returns 403
- [ ] Change a container restart policy in the Health tab — confirm dialog appears and reverts on cancel
- [ ] Verify docker policy refresh no longer uses `_DOCKER_ENRICH_TTL`

### DIFF
--- a/app.py
+++ b/app.py
@@ -38,7 +38,8 @@ VERSION      = "0.24.0"
 DB_PATH      = os.environ.get("DB_PATH", "/data/gpu.db")
 MCP_IDLE_SEC = 45   # seconds without MCP activity before the pill shows idle
 INTERVAL     = int(os.environ.get("SAMPLE_INTERVAL", "10"))
-RETENTION    = int(os.environ.get("RETENTION_DAYS", "180")) * 86400
+_RETENTION_DAYS_DEFAULT = int(os.environ.get("RETENTION_DAYS", "180"))
+RETENTION    = _RETENTION_DAYS_DEFAULT * 86400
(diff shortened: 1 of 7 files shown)


SikamikanikoBG:
Thanks @pehota — merged! Configurable data retention has been on the wishlist for a while now, and folding the hardening fixes in alongside it is exactly the kind of contribution that keeps the project healthy. Appreciate it.

Issue #219, opened 2026-06-29
🧠 AI model catalog — see available models across providers (Ollama, vLLM, …) at a glance


SikamikanikoBG:
Right now the dashboard tells me which models are *loaded in VRAM* right this second. What I keep wanting is the other half: what models do I actually *have* available on each box, across each provider — without SSHing in to run `ollama list` or poke a vLLM endpoint.

**Ask:** a single "AI models" view that catalogs the *available* models per host, grouped by provider.

**Providers to start with**
- Ollama (`/api/tags`)
- vLLM / OpenAI-compatible servers (`/v1/models`)
- llama.cpp server, LM Studio, etc. as follow-ups

**Per model, show at a glance**
- Name / tag
- Provider + which host it lives on
- Size on disk (and quant where we can get it)
- Loaded right now? (tie into the existing live VRAM view)

**Nice-to-haves**
- Dedupe the same model pulled on multiple hosts
- Filter/search across the fleet
- Surface it through the MCP + `/metrics` too, not just the UI

The goal: one place for "what can I run, and where" — no leaving the dashboard.

hiSandog:
For the AI model catalog, provider probing should probably be cached and failure-tolerant. A host with an unreachable Ollama/vLLM endpoint should show stale/unknown status without blocking the whole dashboard. Capturing provider, host, model name, disk size, quantization when available, and loaded-now state would line up well with the existing VRAM-focused view.

1HazyOne707:
I'd like to take this on.

The AI model catalog fits well with the existing GPU/model monitoring direction. I'll look at adding provider discovery in a way that is failure-tolerant (unreachable Ollama/vLLM endpoints shouldn't block the dashboard), with cached probing and graceful unknown/stale states.

I'll start by mapping the existing GPU monitoring/model discovery code so this integrates cleanly rather than adding a parallel path.

SikamikanikoBG:
Awesome — consider it yours. GitHub won't let me formally assign non-collaborators, so nothing to wait on: just open a PR with `Closes #219` when you're ready.

Love the failure-tolerant angle — keeping an unreachable Ollama/vLLM endpoint from blocking the dashboard, with cached probing and graceful unknown/stale states, is exactly the right instinct here. And thanks for mapping the existing GPU/model-discovery path first rather than adding a parallel one; that'll keep it clean.

Shout if anything in there is unclear.

SikamikanikoBG:
[...]

PR #236, opened 2026-07-14
feat: make AI model registry fleet aware

1HazyOne707:
## Summary

First step of making the AI model registry fleet-aware (#219 follow-up). This PR lays the groundwork on the merge/dedup and UI side; it does **not** yet pull real data from remote hosts — that's the next commit, in progress.

### Backend
- Fixed a thundering-herd cache race in `_model_registry()` (double-checked locking — a burst of concurrent callers hitting a stale cache no longer triggers N parallel fetches to ollama).
- `_merge_registry()`'s dedup key is now `(name, provider, host)` instead of `(name, provider)`, so the same model on two different hosts won't collapse into one entry once host data actually flows in.

### Dashboard
- Registry card no longer hides on remote-host tabs.
- Rendering now groups by host, then provider.

### Tests
- 94 passing, including new coverage for the host-aware dedup key.

## What's NOT done yet (next commit)

Right now every `model_catalog` entry is tagged with `socket.gethostname()` inside `sample_once()` — but that function only ever runs on the hub, so every entry still reports the hub's own hostname. `/api/models` also still only reads the hub's local `LATEST`, not `HOST_DATA` (the fleet cache other host-wide features already use).

So today this PR makes the *plumbing* host-aware, but doesn't collect anything from remote hosts yet. Real fleet-wide data requires:
1. Extending `probe.py` (the script pushed to remote hosts over SSH) with container + AI-provider discovery — it currently only reports system metrics, no model detection at all.
2. Folding that into `probe_host_metrics`'s returned JSON so it lands in `HOST_DATA[name]["data"]["model_catalog"]`.
3. Updating `/api/models` to aggregate local + every host's `HOST_DATA` entry (mirroring the pattern `_iter_fleet_hosts()` already uses for `os_updates_summary()`).

Wanted to be upfront about the actual state rather than let the title overclaim — pushing the remaining pieces next.


1HazyOne707:
Following up — the remaining pieces from the earlier note are now in:

1. `probe.py` (the script pushed to and run on every registered remote host over SSH) now has `read_ollama_models()`, polling that host's own ollama (`/api/ps`, falling back to `/api/tags`) and reporting it under `model_catalog` in the probe's JSON output.
2. `/api/models` (`backend/api/gpu.py`) now aggregates the hub's local catalog with every registered host's `HOST_DATA[name]['data']['model_catalog']` before merging — no extra I/O, it rides the existing host-poll cycle.

Verified end-to-end: ran `probe.py` directly and confirmed it correctly detects and reports real local ollama models tagged with the actual hostname.

Scope: Ollama only for now, matching the issue's own phased provider list. The other 30+ providers already supported locally (vLLM, llama.cpp, LM Studio, ComfyUI, …) would need the same treatment in `probe.py` — tracked as a natural follow-up rather than bundled into this PR.

23/23 registry tests passing. Full suite: 487 passed, 5 failed — those 5 are an unrelated, already-diagnosed issue on a separate branch (fix pushed, PR not yet opened pending a reply elsewhere), present on `main` before this commit.

SikamikanikoBG:
Merged — thank you @1HazyOne707! Two notes on how it landed:

1. I retargeted the base from `main` to `next` before merging — all feature PRs go through `next` (CONTRIBUTING.md, "Where to branch from"); `main` only moves at release time. No action needed, just flagging for next time.
2. Reviewed the full stack: the double-checked locking in `_model_registry()` is a genuine catch (the fetch happens outside the lock, so the stampede window was real), the `(name, provider, host)` dedup key + tests are exactly right, and the probe-side `read_ollama_models()` with the `/api/ps` → `/api/tags` fallback rides the existing poll cycle with no extra I/O, as advertised.

Also appreciated the honest "what's NOT done yet" note in the description — and then actually delivering the missing pieces. That made this easy to review. The remaining providers (vLLM, llama.cpp, …) as a probe-side follow-up sounds right whenever you feel like it.

PR #237, opened 2026-07-15
fix(gpu): report GTT, not the 512 MB VRAM carve-out, on unified-memory APUs

andreahaku:
## What

On unified-memory APUs (Ryzen AI Max+ 395 / Strix Halo, Radeon 8060S), the GPU
panels read the dedicated-VRAM carve-out instead of the memory the GPU actually
uses. `amdgpu` exposes a tiny fixed VRAM window (512 MB here, the BIOS minimum)
via `mem_info_vram_*`, while models and buffers live in **GTT**: system RAM
mapped to the GPU, up to ~124 GB on a 128 GB box.

Caught live on an idle box: the GPU tile read **148 MB / 512 MB = 29% full**,
and the insight feed permanently flagged **"GPU VRAM ran low, only 364 MB
free"**, because `free_min = mem_total - peak` is always below
`PRESSURE_FREE_MB` (2048) when `mem_total` is 512 MB. Meanwhile the real pool
(`mem_info_gtt_total`) read 124 GB with ~0 used.

## Fix

`amd_gpus()` now reads `mem_info_gtt_total/used` and reports GTT when the card
looks like an APU (VRAM <= 1 GiB alongside a large GTT pool). Discrete cards
(large dedicated VRAM) fail that test and keep reporting VRAM exactly as before,
so there is no behaviour change for them.

`app.py` `amd_gpus()` only. After the change the same idle box reads **18 MB /
124.0 GB, 0% full**, and the insight flips to "GPU has plenty of headroom". When
a model loads into GTT, residency and the pressure math track the real number.

### DIFF
--- a/app.py
+++ b/app.py
@@ -4795,8 +4795,21 @@ def amd_gpus(drm_root=None):
                     continue
         except OSError:
             continue
-        total  = _amd_read_int(os.path.join(dev, "mem_info_vram_total"))   # bytes
-        used   = _amd_read_int(os.path.join(dev, "mem_info_vram_used"))    # bytes
+        vram_total = _amd_read_int(os.path.join(dev, "mem_info_vram_total"))  # bytes
+        vram_used  = _amd_read_int(os.path.join(dev, "mem_info_vram_used"))   # bytes
+        # APU / unified-memory iGPU (e.g. Ryzen AI Max / Strix Halo): the dedicated
+        # VRAM is a tiny BIOS carve-out (<= ~1 GiB) while the real working set — where
+        # models actually load — lives in GTT (system RAM mapped to the GPU, up to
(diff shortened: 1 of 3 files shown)


SikamikanikoBG:
Andrea, this is a great catch — thank you. The diagnosis is exactly right, and tracing the stuck "VRAM ran low" insight back to the free_min math is the kind of detail that makes a fix trustworthy.

One addition on top: the same reader exists twice — `app.amd_gpus()` (hub-local) and `probe._amd_gpu_sysfs()` (the remote probe). On a hub+probe setup the probe is how an APU box usually reports in, so I mirrored your GTT heuristic there too and added tests for both readers (GTT reported over the carve-out, missing `gtt_used` degrades to 0, discrete cards unaffected). All green on 3.13.

Merging now. Really appreciate the contribution — hope to see more.

PR #247, opened 2026-07-25
feat(gpu): per-process VRAM attribution on AMD, from the DRM fdinfo nodes

andreahaku:
Follow-up to the note left in the collector: NVIDIA cards get per-process attribution from `nvidia-smi --query-compute-apps`, AMD cards had none — so on an AMD box "VRAM by service" is a single anonymous `system/other` block and the Containers tab can't say which service is holding the memory.

amdgpu has no compute-apps equivalent, but the kernel already publishes what's needed: every fd open on a DRM node carries its client's residency in `/proc/<pid>/fdinfo/<fd>` — `drm-resident-<region>` (or `drm-memory-<region>` on kernels from before that was standardised), alongside `drm-pdev` and `drm-client-id`. Reading those yields the same `(pid, MB)` pairs the NVIDIA path already feeds to `service_for_pid()`, so both vendors land in the same per-service totals with no changes downstream.

No new dependency, no new configuration, no subprocess.

### Measured

On a Ryzen AI Max (Strix Halo, 124 GiB GTT) under load, against the kernel's own `mem_info_gtt_used`:

| | |
|---|---|
| attributed | 24160 MB of 24174 MB reported by the kernel — **99.9%** (remainder is scanout) |
| scan cost | **8 ms** across 550 processes / 1621 fds |

The GPU tab goes from one grey `system/other` band to a row per container, and "Services on the GPU" fills in.

### Details that turned out to matter

- **A bare number is bytes, not KiB.** The fdinfo formatter only scales up while the value divides evenly by 1024, so an unaligned figure prints raw — which is why a zero always reads `0` and never `0 KiB`. Reading bare values as KiB overcounts them 1024x.
- **Clients are deduped by `(drm-pdev, drm-client-id)`.** dup()'d fds, threads and forked children all republish the same client; two cards can legitimately hand out the same id, hence the pdev in the key. Processes are walked lowest-PID-first, so a client visible through several of them (a supervisor holding an inherited fd) always lands on the same service instead of flipping between samples.
- **`drm-shared-<region>` is subtracted.** A buffer shared between two clients counts in the residency of both, and their client-ids differ, so dedup can't catch it — crediting it whole to each would push per-service totals past the card's capacity.
- **An fd is recognised by its device number, not its path.** A container can map the render node anywhere (`--device=/dev/dri/renderD128:/dev/gpu0`), and a path test would silently drop all of that container's memory.
- **A stat we're not allowed to make means "read fdinfo anyway", never "skip".** A process in another user namespace (rootless podman) denies it while leaving fdinfo readable — and that container is often the one holding all the VRAM.
- **On an APU the GTT field is read**, for the same reason `amd_gpus()` reports GTT there. That threshold now lives in one shared helper, so the panel and the attribution can't disagree about a card.

### Degrades quietly

No AMD card, no `drm-pdev` to match a card by, or a kernel with neither key → empty list, and the host keeps today's behaviour: GPU panel, no per-service attribution. Nothing raises.

### Two consequences beyond the GPU panel

Since `gpu_pids` and `procs` feed them: an AMD-only training job now appears in the Training and DevTools panels, and an AMD-only container enters the OOM scan's target set. Both look desirable, but flagging them since they're outside what the title suggests.

### Known limit

This sees DRM clients. ROCm processes are covered in practice — they hold a render-node fd, and a PyTorch ROCm container is attributed correctly in the numbers above — but a process allocating solely through `/dev/kfd` with no DRM fd would not be.

Also hub-only: `probe.py` doesn't do per-process attribution for NVIDIA either, so remote hosts are unchanged and this stays symmetric with the existing behaviour.

### Tests

`tests/test_amd_gpu.py` grows a `TestAmdPerProcess` class over a fake `/sys/class/drm` + `/proc`, covering the APU/discrete field choice, the legacy keys, shared buffers, same-client-across-fds and across-PIDs, two cards sharing a client-id, a device mapped to another path, a process dying mid-scan, an unreadable fd in another user namespace, unit parsing, and the degraded paths. 35 pass in that file.

Only `probe.py`, docs and version files are untouched — happy to adjust anything that doesn't match how you'd like this shaped.


SikamikanikoBG:
@andreahaku First: this is an outstanding PR — the fdinfo detail work (bytes-vs-KiB, drm-shared subtraction, device-number matching, the rootless-podman case) is the deepest anyone has gone on this codebase's GPU side. Which makes what I have to say awkward, and it's on us, not you.

The conflicts you see are because #249 — our own implementation of the same feature, same DRM-fdinfo approach — merged on 2026-07-27, two days after you opened this. It came out of a Discord report against v0.26.0, and we failed to connect it with this open PR before merging.

That said, your version does several things the merged one doesn't, and I'd genuinely like them in the tree:

- reads the standardized `drm-resident-*` keys (the merged code only reads `drm-memory-*`/`drm-total-*`)
- subtracts `drm-shared-*`, so cross-client shared buffers aren't double-credited
- recognises the render node by device number, not path — a `--device=/dev/dri/renderD128:/dev/gpu0` remap silently drops that container's memory in the merged version
- treats a denied stat as "read fdinfo anyway" (the rootless-podman container is often the one holding all the VRAM)
- parses bare fdinfo values as bytes, not KiB
- a far deeper test suite (35 cases over a fake /sys + /proc)

If you're up for it: a rebase onto current `next` as a hardening PR against `amd_fdinfo_procs()` carrying those behaviors — that merges. If you'd rather not redo work we made you redo, say so and we'll port your fixes ourselves with `Co-authored-by` credit. Either way, thank you — and sorry again for the collision.

andreahaku:
No hard feelings, these things happen. If anything #249 landing first made this a better patch. Your per-pdev split with the `unified` flag is finer than my host-wide APU heuristic was, so your architecture and the collector wiring stay as they are. Mine is now just the hardening pass you suggested.

Inside `amd_fdinfo_procs()`: resident keys win, with `drm-memory-*` and then `drm-total-*` as fallbacks, picked by key presence rather than truthiness. An evicted client can legitimately report 0 resident, and a truthiness chain would credit it with its total. `drm-shared-*` gets subtracted. Fds are matched by device number, and a denied stat means read fdinfo anyway, which is the rootless podman case. Bare values parse as bytes.

One thing I carried that wasn't on your list: the (device, client-id) dedup is now global and lowest PID first, not per pid. A supervisor holding an inherited fd was counting the worker's buffers twice, once per service. And readdir order could flip the owner between samples. When `drm-pdev` is missing the dedup keys on the fd's device number, then on the pid, so pdev-less multi-card kernels don't drop a client. If you'd rather keep the per-pid semantics I'll drop that part.

Tested on my Strix Halo box, 128GB unified. Two llama-server instances in rootless podman come out at 22.5GB and 2.4GB, against 24.6GB from sysfs. The 1% overshoot is per-client resident pages vs the card-level counter, they just don't account identically. The test file is at 42 cases now, on real symlinks, so fd identity actually goes through os.stat on both roads: device present, and stat failing like on GPU-less CI.

Unrelated: `tests/test_snapshots.py` has a metrics snapshot mismatch on current next. I reproduced it on a clean checkout of 327e6b5, so if CI goes red there it's not this branch.

It's been fun going this deep into the GPU side of the codebase, so really, no problem. I run this daily on that box and I'll keep sending patches as I trip over things.

SikamikanikoBG:
Merged — thank you. The fdinfo route is the right call for this project (sysfs-only AMD back-end, no ROCm dependency), and the details you chased are the ones that would have quietly produced wrong numbers for months: the bare-number-is-bytes formatter quirk, keying fd identity on the device number so a remapped render node inside a container doesn't lose its attribution, and subtracting `drm-shared-*` so a dma-buf isn't counted against two clients at once. The lowest-PID-first dedup is a nice touch too — a supervisor holding an inherited fd would otherwise have made the per-service history sawtooth between samples.

99.9% attributed in 8 ms across 1621 fds is a good number to have measured.

PR #248, opened 2026-07-26
feat(metrics): export RAPL CPU/DRAM package power via Prometheus

pehota:
## Summary

Adds real-time CPU and DRAM power consumption monitoring for Intel/AMD hosts that support RAPL, giving users visibility into per-package wattage directly on the dashboard.

- Reads RAPL energy counters from `/sys/class/powercap` and exposes CPU and DRAM package power as Prometheus gauges
- Mounts `/sys/class/powercap` in the Docker container so RAPL data is accessible inside the container
- Fixes broken RAPL tests and guards `LATEST` against transient `None` values

## Test plan

- [ ] RAPL metrics appear in Prometheus output on a host with RAPL support
- [ ] Container starts correctly with the new `/sys/class/powercap` mount
- [ ] RAPL tests pass (`pytest tests/`)
- [ ] No regression on hosts without RAPL (metrics absent, no errors)

### DIFF
--- a/app.py
+++ b/app.py
@@ -145,6 +145,8 @@ if _PROM_OK:
         "gpu_temp":          _make_gauge("homelab_gpu_temp_c",          "GPU temperature (°C)",              ["gpu"]),
         "gpu_power":         _make_gauge("homelab_gpu_power_w",         "GPU power draw (W)",                ["gpu"]),
         "host_cpu":          _make_gauge("homelab_host_cpu_pct",        "Host CPU usage (%)"),
+        "host_cpu_power":    _make_gauge("homelab_host_cpu_power_w",    "Host CPU package power, RAPL (W)"),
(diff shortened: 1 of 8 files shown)


pehota:
@claude what's up? Care to do a review?

SikamikanikoBG:
Merged — thanks @pehota! The CI review job skipped on this one, so here's the review pass it owed you:

- Checked the `cpu_w`/`dram_w` → `cpu_power`/`dram_power` rename against every consumer — the collector in `backend/collectors/__init__.py` was the only reader and it's updated here, so nothing else breaks.
- Dropping the `endswith(":dram")` match is right: powercap `name` files never contain a colon, so that branch was dead code and the tests now test reality.
- The probe-side two-read dwell is the correct trick for a one-shot process, and counter wraparound is handled via `max_energy_range_uj`.
- Keeping the last good value in `LATEST` when a cycle returns `{}` is a nice touch — no flapping gauges.
- All three compose mounts are `:ro`.

In `next` now — ships with the next minor.

PR #254, opened 2026-07-28
feat(gpu): AMD parity for the GPU panel — clocks, perf level, caps, and real card names

andreahaku:
Follow-up to #249's AMD side, aiming at panel parity with NVIDIA on AMD hosts. Two things, plus one honesty fix that rode along.

**gpu_extra on AMD.** amd_gpus now attaches the fields _enrich_gpus produces, read from amdgpu sysfs in the same per-card pass: clk_sm/clk_mem from the starred pp_dpm_sclk/mclk rows (label-matched hwmon freq fallback), mem_util from mem_busy_percent, power_limit from hwmon power1_cap, pstate from power_dpm_force_performance_level, temp_mem from the hwmon channel labelled `mem`. Every node is optional: missing file, no field, chip stays hidden. The aggregate is now built across vendors, which also fixes a pre-existing hybrid quirk: pooled draw was compared against the NVIDIA-only cap, so the power chip could read over 100% as soon as an AMD card drew anything. The UI distinguishes AMD perf levels (auto/low/high/manual, a policy you set) from NVIDIA numeric P-states, each with its own tooltip.

**Real card names.** APU kernels expose no product_name, so a Radeon 8060S showed as "AMD GPU 1". When product_name is absent the name now comes from the host's pci.ids through the HOST_ROOT mount (hwdata and misc locations), looked up in the 0x1002 block. The bracket in an AMD entry is used when it names one retail product; when it lists several variants sharing the silicon the codename is kept, since sysfs can't tell an 8050S from an 8060S.

**The honesty fix:** _gpu_extra emitted mem_util=0 for cards that never measured it (APUs have no mem_busy_percent), and the UI rendered that as a confident "0% Mem-BW util" chip. The aggregate now omits the field unless some card measured it; a measured 0% still passes through.

Verified live on the same Strix Halo box as #247 (128GB unified memory): the panel now reads "AMD Strix Halo", 600 MHz core clock at idle from the starred dpm row, perf level auto. And the absences are correct too: this APU has no mem_busy_percent, no power1_cap, and doesn't star any pp_dpm_mclk row, so all three of those chips stay hidden instead of showing zeros. 21 new tests over fake sysfs trees and a fake pci.ids.

Scope note: this covers the hub host. probe.py still reports the fallback name and no extras for remote AMD hosts; happy to port the same two lookups there as a follow-up. Same for gpu_metrics parsing (throttle reasons, extra temps), which looked like its own PR.


SikamikanikoBG:
Merged — thank you, and thanks for the pci.ids work in particular; "AMD GPU 1" for an APU always looked broken and the codename-vs-retail-bracket handling is more careful than I'd have bothered with.

The `_gpu_extra` fix is the one I'm happiest about — pooled draw against the NVIDIA-only cap meant the power chip read over 100% the moment an AMD card drew anything, and publishing the cap only when every card reports one is the honest answer rather than papering over it. Same for leaving `mem_util` absent instead of emitting a confident 0% on cards that never measured it.

One heads-up on process: this had to be merged with a merge commit resolving against #247, since your new helpers landed exactly where #247 rewrote `_fdinfo_kib` into `_fdinfo_bytes`. Purely adjacent, nothing semantic — I pushed the resolution to your branch so CI could verify it before it hit `next`.

PR #261, opened 2026-08-01
feat(gpu): unified GPU dashboard across all hosts — per-card history, fan speed, and thermal alerts

SikamikanikoBG:
## What this does

The GPU tab showed detailed charts for the hub but only a static snapshot for
remote machines. This makes the interface consistent everywhere and adds what
you couldn't see at all: fan speed, per-card history, and which card each
service is actually using.

Completes the v0.27.0 roadmap item: *"Per-host GPU history and service
attribution need per-host storage — that's the next slice."*

## Why this was needed

Not a UI limitation — a storage one. `host_samples` pooled a box's GPUs into one
row and dropped GPU temperature entirely, so remotes had nothing to chart.
`renderRemoteGpu()` said as much in its own caption.

The fix changes the schema: `gpu_samples` and `proc` gain a `host` column, the
hub writes itself as `host='local'`, and one reader serves every machine. That
removes `renderRemoteGpu()`, the `tabId === 'gpu'` case in
`renderLocalOnlyNotice()`, and the GPU branch of `buildCharts()` — deleted, not
extended.

## What's new

- **Fan speed** — never collected anywhere in this codebase before: NVIDIA
  `fan.speed`, AMD `pwm1`/`fan1_input`.
- **Full telemetry on remotes** — the probe queried 7 nvidia-smi fields while the
  hub queried 14. Remotes now get the same set: memory bandwidth, clocks, power
  cap, memory-junction temperature, perf state and throttle reasons.
- **Per-card service attribution** — `--query-compute-apps` identifies a process's
  card by UUID and nothing was asking for it. Mapping UUID → index and pid →
  cgroup shows ollama's 63 GB split as 22.5 / 22.1 / 18.8 GB across the cards
  instead of one pooled number.
- **Side-by-side card panels** with a shared y-scale per metric, so a taller
  temperature line really is a hotter card rather than an autoscaling artefact.
  Throttle periods are shaded on the sparkline itself. A by-metric view
  transposes the same data when the question is "which card is the problem".
- **Card-level alerts** — sustained rather than instantaneous, hysteresis on
  clear, per card and per host. Power caps never trigger one.

## Key design choices

- **Temperature is MAX across cards, not the average.** An average over one card
  at 87 °C and two idling at 45 °C reads as 59 °C and hides the problem.
- **Throttle counting excludes `SW_POWER_CAP`.** A deliberately power-limited box
  sits at its cap by design and shouldn't read as permanently throttled.
- **Attributed power is labelled an estimate everywhere it appears.** GPUs meter
  per card, never per process, so it's apportioned by VRAM share above the idle
  floor — and the idle floor is its own band rather than billed to whoever is
  resident.
- **Unsupported metrics stay absent, never 0.** A passively cooled card and a
  stalled fan are different claims, and one of them raises an alert.

## Testing

738 passing, 63 new. The 6 failures on this branch are the same 6 that already
fail on `next` (`test_public_status` maintenance flags, `test_no_silent_swallow`).

Migration verified against a copy of the live 290 MB database: all 39,046
`gpu_samples` and 47,896 `proc` rows preserved and attributed to `local`.

Verified live on a 3×3090 host. Seven bugs turned up only by doing that — a
retired GPU reading as a hardware failure, a zero-RPM idle fan reading as a
stalled fan, and every sparkline rendering invisible because `.spark` collided
with an existing class. Details in `design/gpu-cockpit/journal.md`.

## Note

`test_db_factory.py::test_different_threads_return_different_conns` is flaky on
`next` independently of this branch — roughly 1 in 3 full-suite runs. Not
touched here.


qrakhen:
Wait, so you are vibe-coding your PRs AND let those PRs be reviewed by... an AI?
And I was sitting there, wondering why my PRs into your repos are not getting accepted :/

PR #265, opened 2026-08-07
fix(containers): a nested bind mount was billed to the one container that named it

### LINKED ISSUE #264: Container disk: a nested bind mount is billed to one container when a parent path makes it shared
## What happened

On the Containers tab, a container that has been **stopped for two weeks** is reported as using
**1.34 TB** of disk. Its writable layer is 40.9 kB.

The number is the whole of `/srv/models`, a host data directory it bind-mounts read-write. That
directory is also reachable read-write from five *other* containers, because they are
toolbox/distrobox containers and toolbox bind-mounts the host root at `/run/host`
(`Source: "/", RW: true`).

`_shared_mount_sources()` is meant to catch exactly this — its docstring says shared infrastructure
"shouldn't be attributed to any single container" — but it compares mount sources by **string
equality**. No one else mounts the literal string `/srv/models`, so the directory looks exclusive
and the entire tree is billed to the one container that named it.

The row then dominates the tab's total, and since a stopped container can't be "using" a terabyte it
reads as a display error rather than a measurement choice.

## What you expected

Data that more than one container can write shouldn't be charged to any of them — including when the
sharing happens through a **parent** path rather than the same path. In the case above the container
should fall back to its writable layer (40.9 kB).

## How to reproduce

No GPU or special hardware needed — two containers are enough.

1. Create a directory with some data in it, e.g. `/srv/share`.
2. Start container A with only that directory mounted read-write:
   `docker run -d --name a -v /srv/share:/data alpine sleep infinity`
3. Start container B with the **host root** mounted read-write (what toolbox/distrobox do):
   `docker run -d --name b -v /:/run/host alpine sleep infinity`
4. Wait for the disk figures to refresh and look at the Containers tab.

**Observed:** container A's disk is the full size of `/srv/share` (and B's is the whole host).
**Expected:** both are treated as shared data, so neither is charged for it.

The same happens with any nesting, e.g. A mounts `/srv/share/models` and B mounts `/srv/share` —
different strings, same bytes.

## Environment

- HomeLab Monitor version: **0.29.0**
- Host OS: **Fedora Linux 43 (Server Edition)**, kernel 7.1.5
- Container engine: **Podman 5.8.4** (rootful monitor reading the rootless socket),
  podman-compose 1.5.0
- GPU + driver: AMD Strix Halo (unified memory) — not relevant to this bug
- Browser: not a UI bug; `/api/health` returns the same value

## Logs

Nothing is logged — the collector doesn't fail, it computes a number that is arithmetically correct
under the current rule. The relevant values instead:

<details>
<summary>measurements from the affected host</summary>

```
# the container reported at 1.34 TB, stopped 2 weeks ago
$ podman ps -a --size --format '{{.Names}} | RW={{.Size}}' | grep llama-rocm
llama-rocm | RW=40.9kB (virtual 8.04GB)

# its only large read-write mount
$ podman inspect llama-rocm --format '{{range .Mounts}}{{.Type}} RW={{.RW}} SRC={{.Source}}
{{end}}'
bind RW=true SRC=/home/andreahaku
bind RW=true SRC=/dev
bind RW=true SRC=/srv/models

$ du -sb /srv/models
1251228588133   /srv/models

# five other containers can write the same tree through the host root
$ for c in $(podman ps -aq); do
    podman inspect $c --format '{{.Name}} {{range .Mounts}}{{if and .RW (eq .Source "/")}}mounts / RW{{end}}{{end}}'
  done | grep 'mounts / RW'
llm-finetuning mounts / RW
ds4-rocm-7.2.4 mounts / RW
unsloth-studio mounts / RW
llama-vulkan-new mounts / RW
llama-vulkan-radv mounts / RW
```

</details>

## Anything else

Happy to send a PR — the change I have in mind is small and stays inside the existing intent: decide
sharing by path **coverage** instead of string equality, so a source counts as shared when another
container mounts it, an ancestor of it, or a descendant of it. Comparison per path component, so
`/srv` doesn't "contain" `/srvfoo`.

One consequence worth agreeing on before I open it: on a host where a container mounts `/`
read-write, every other read-write mount becomes shared, so those rows fall back to their writable
layer. That reads as correct under the "belongs to none of them" principle — if a container can
write the whole host, no host data is exclusively anyone's — but it is a visible behaviour change for
toolbox/distrobox users, and it's your call whether that's the trade-off you want. Read-only mounts
(the monitor's own `/rootfs`) are unaffected, so the normal Compose setup with per-service volumes
doesn't change at all.

Two adjacent things I noticed while digging, deliberately **not** part of that change — say the word
if you'd rather have them in the same PR or in issues of their own:

1. A single container that mounts both a parent and a child (`/srv` and `/srv/models`) has the
   nested part counted twice in its own total.
2. `_container_disk()` keeps `max(total, prev)` whenever any mount's `du` overruns its timeout, so a
   figure can grow but never shrink while scans stay incomplete. Deliberate per the docstring, but it
   does mean a stale high-water mark can persist after data is deleted.

andreahaku:
## What this changes

Fixes #264.

`_shared_mount_sources()` decided whether a mount was shared by comparing sources for **string
equality**, so it only ever caught two containers naming the same path. It missed the nested case:
one container mounting `/srv/models` and another mounting `/` (what toolbox and distrobox do, at
`/run/host`) write the same bytes under two different strings, so the whole tree was billed to the
single container that named it. On the host that surfaced this, a container **stopped for two weeks**
with a **40.9 kB** writable layer was reported at **1.25 TB**.

Sharing is now decided by path **coverage**, and deliberately **one-directional**: a source is shared
when another container mounts the same source, or a directory *above* it. Mounting a parent does not
make you lose it because someone else mounted a subdirectory of yours — a container with 500 GB under
`/srv/media` keeps being billed for it when another mounts only `/srv/media/photos`. The nested part
stops being double-counted, the exclusive part doesn't vanish from the report.

Sources are normalised once, in `_norm_src`, and both the sharing check and the skip inside
`_container_disk` go through it — normalising only one side would make `src in shared` miss silently.
That also gives `_shared_mount_sources` the same "absolute paths only" guard `_container_disk` always
had, so podman's pseudo-sources (`devpts`) stop landing in the shared set; they were never measurable
anyway.

Comparison is per path component, so `/srv` doesn't "contain" `/srvfoo`, and compose named volumes
sharing a prefix (`p_db/_data` vs `p_db_data/_data`) don't collide.

## Before / after on the real host

Fed the actual `/containers/json?all=1&size=1` payload from the affected host (13 containers) through
the old and the new code, with `_dir_size` returning the measured `du -sb /srv/models`
(1 251 228 588 133) and 0 elsewhere:

| container | before | after |
|---|---:|---:|
| `llama-rocm` (stopped 2 weeks) | 1 251 228 629 020 | **40 887** |
| the other 12 containers | unchanged | unchanged |

Nothing that was genuinely shared before stops being shared. The one source that leaves the set is
`devpts`, a relative pseudo-source that `_container_disk` already skipped, so no bytes move.

## Checklist

- [x] **Base branch is `next`** (the integration branch), not `main`.
- [x] Branched off the **latest `next`** — re-checked with `git fetch` just before opening this;
      `next` hasn't moved, so no rebase was needed.
- [ ] Tested locally with `docker compose up -d --build` — **not** done that way, and I'd rather say
      so than tick it. My deployment is rootful Podman and I can't rebuild it from this account.
      Instead I ran the collector against the live container payload of the affected host (table
      above), which I think is stronger evidence than a dashboard glance. Happy to do the Compose run
      too if you want it before merging.
- [x] No version bump in this PR.
- [ ] For a new monitor / probe: followed the **"add a monitor" pattern**. *(Not applicable — a fix
      inside an existing collector.)*

## Notes for the reviewer

**The trade-off to agree on.** On a host where some container mounts `/` read-write, every other
read-write mount becomes shared, so those rows fall back to their writable layer. That's the honest
reading of the "belongs to none of them" rule already in the docstring — if a container can write the
whole host, no host data is exclusively anyone's — but it can empty the disk column for a whole host,
and doing that silently would be impossible to diagnose in the field. So the collector now prints one
`NOTE:` line when the covered set changes, naming up to three sources. If you'd rather keep whole-root
mounts out of the coverage rule entirely, or drop the log line, say so and I'll rework it.

**Known limit, documented rather than fixed.** `_covers` is lexical and can't see through symlinks:
if `/srv/models` is a link to `/mnt/disk/models` and another container mounts `/mnt/disk`, the sharing
goes unnoticed. Resolving it means `realpath` against the host rootfs, which isn't always mounted, so
the docstring states the limit instead. This is pre-existing behaviour, not a regression.

**Tests.** `tests/test_container_disk.py` is new: 25 cases over `_norm_src`, `_covers` and
`_shared_mount_sources`, plus four end-to-end through `_container_disk` with `_dir_size` stubbed. Two
of them earn their keep specifically:

- `test_without_the_shared_set_it_is_billed` is the control for the fix — without it the main test
  would pass even if the filter did nothing.
- `test_parent_owner_still_billed_when_a_child_is_shared` pins the one-directional choice, so if
  someone later makes the rule symmetric the test says what breaks and why.

They run under plain `unittest` too. Locally, mirroring CI: `python -m py_compile app.py probe.py`
clean, `pytest tests/test_snapshots.py` 71 passed, `pytest tests/test_container_disk.py` 25 passed. I
noticed CI's pytest step is scoped to `tests/test_snapshots.py`; I left the workflow alone rather than
widen it inside a fix PR, so say the word if you'd like these wired in.

**Two adjacent things I deliberately left alone** (also in the issue, happy to take either separately):

1. A single container mounting both a parent and a child (`/srv` and `/srv/models`) counts the nested
   part twice in its own total. Untouched here — it's one container's own arithmetic, not sharing.
2. `_container_disk()` keeps `max(total, prev)` whenever any mount's `du` overruns its timeout, so a
   figure can grow but never shrink while scans stay incomplete. Deliberate per the docstring, but it
   does let a stale high-water mark outlive deleted data — which is why the dashboard was showing
   1.34 TB for a tree that measures 1.25 TB today.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1795,10 +1795,11 @@ def _container_disk(ct, prev=None, shared=frozenset()):
     """A container's real on-disk footprint: writable layer (SizeRw) + every
     read-write volume and bind mount it owns. Read-only mounts (configs, the
     docker socket, our own /rootfs) aren't this container's data, so they're
-    skipped, and so are sources in `shared` — a bind/volume mounted by more than
-    one container (e.g. a common /srv/share) belongs to none of them, and counting
-    it under each both double-counts and makes unrelated rows show the same huge
-    number. Falls back toward SizeRw alone when host paths can't be measured —
+    skipped, and so are sources in `shared` — data reachable read-write from more
+    than one container (e.g. a common /srv/share, or a directory another container
+    covers by mounting its parent) belongs to none of them, and counting it under
+    each both double-counts and makes unrelated rows show the same huge number.
+    Falls back toward SizeRw alone when host paths can't be measured —
     e.g. HOST_ROOT not mounted — which is the old behaviour.
 
     `prev` is this container's last computed total: if a mount's `du` overruns
@@ -1809,8 +1810,8 @@ def _container_disk(ct, prev=None, shared=frozenset()):
     for m in (ct.get("Mounts") or []):
         if not m.get("RW") or m.get("Type") not in ("volume", "bind"):
             continue
-        src = m.get("Source")
-        if not src or not src.startswith("/") or src in shared:
+        src = _norm_src(m.get("Source"))
+        if not src or src in shared:
             continue
         sz = _dir_size(src)
         if sz is None:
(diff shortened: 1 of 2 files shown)


SikamikanikoBG:
Merged — thank you. This is your fourth PR here, and the first that came as a complete package: issue diagnosis, fix, and real-world before/after data. Seeing the 1.25 TB become 40.9 kB in your table meant I could take this without owning a toolbox host.

Three specific points where your approach was spot-on:

1. **The one-directional rule.** The symmetric version is the obvious one to write, and it trades one wrong number for another — I'd probably have written it and found out later, on my own media host. Good call.
2. **Normalising in `_norm_src` rather than only where sharing is computed.** That's the kind of thing that never shows up in a diff review as a bug, only as a silence six months on.
3. **Honesty over checkboxes.** You wrote "I'd rather say so than tick it" about the Compose box. Running the collector against the affected host's real payload is stronger evidence than a dashboard glance anyway — but the plain "I can't do this and here's why" is worth more to me than a ticked box, and I'd rather the next contributor knew that.

On the trade-off you flagged: accepted. A container with read-write `/` really does mean no host data is exclusively anyone's, and the `NOTE:` line is what makes that diagnosable instead of mystifying. Keep it.

Two follow-ups I'd take separately if you want them, no obligation on either:

- Wiring `tests/` into CI so `test_container_disk.py` actually runs there — right now the pytest step is still scoped to `tests/test_snapshots.py`, so your 25 cases sit in the repo unexecuted by the pipeline.
- The `max(total, prev)` high-water mark you mentioned. Deliberate per the docstring, but as you spotted, it lets a stale high mark outlive deleted data — which is why that host read 1.34 TB for a tree that measures 1.25 TB today.

Shipped as [v0.29.1](https://github.com/SikamikanikoBG/homelab-monitor/releases/tag/v0.29.1).

PR #272, opened 2026-08-11
feat(dashboard): Reduced motion setting to disable gauge animations

pehota:
## Summary
- Adds a "Reduced motion" toggle (Settings → General, `reduced_motion` setting, default off) that turns off the Overview cockpit's gauge/count-up/meter-bar animations — gauges just show the current value. Also respects OS-level `prefers-reduced-motion` automatically.
- Adds a global CSS kill switch (`html[data-reduced-motion="1"] * { transition:none!important; animation:none!important }`) so any future CSS-based animation is covered automatically, no per-element opt-out needed.
- Documents the convention in `docs/animations.md` — CSS is auto-covered; JS-driven number tweens must go through the existing `mcCountUp()` helper rather than a new `requestAnimationFrame` loop — and links it from `CONTRIBUTING.md` and a new `CLAUDE.md`.

## Test plan
- [x] Verified in-browser (chrome-devtools): checkbox appears in Settings → General, toggling flips `MC_RM` and the `data-reduced-motion` attribute instantly, persists via autosave, survives reload, and default (off) preserves the original animated behavior.
- [x] Confirmed the three tri-gauge arcs render with `transition: none` (computed style) when enabled, and the normal `1.1s cubic-bezier` transition when disabled.
- [x] `app.py` parses; all inline `<script>` blocks in `dashboard.html` parse.
- [x] Booted the app locally on this branch (based on `next`) and spot-checked Overview, Costs, Containers tabs — no new console errors (only a pre-existing 404 for vendored JS outside Docker, which self-heals via the existing CDN fallback).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,13 @@
+# CLAUDE.md
+
+Project-specific instructions for AI agents working in this repo. See
+[`CONTRIBUTING.md`](CONTRIBUTING.md) for the human-facing contributor guide
+(branching, the "add a monitor" pattern, style) — it applies here too.
+
+## Conventions
+
+- **Animations**: see [`docs/animations.md`](docs/animations.md) before
+  adding any transition, `@keyframes`, or JS-driven value tween. CSS-based
+  animations are covered automatically by the reduced-motion kill switch;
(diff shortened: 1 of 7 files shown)


pehota:
@claude , dude, check my PR

PR #289, opened 2026-08-29
test(dashboard): cover locale re-render paths

### LINKED ISSUE #284: Extend the vm-lift JS test pattern beyond the refresh loop
The dashboard is one ~10k-line HTML file with no build step, so there's nothing to `import` in a test. `tests/js/test_refresh_loop.js` solves this cleanly: it *lifts* the functions under test straight out of `static/dashboard.html` and runs them in a `vm` context against a fake clock — the shipped source is never copied or patched, so the test can only pass by the real code behaving.

It's the only JS test, and the most recent fix work landed squarely in this territory — #273 (chart-refresh cadence) and #274 (per-card panels refreshing in place) were both client-side refresh bugs the harness-style test is built to catch. The pattern has earned its keep; it just isn't used yet.

**Suggested scope**

- Pick one under-tested client area — the realtime tail, the chart-repaint/slide logic, or the locale re-render — and add a `vm`-lift test file in `tests/js/` following the existing harness (wire it into the CI job that already runs `node tests/js/test_refresh_loop.js`).
- The fake-clock + fake-`fetch` scaffolding in the existing file is reusable as-is.

Jokasa7:
## What this changes

Adds a VM-lifted Node test for the dashboard's `lang:changed` listener and runs it in both the integration and release validation workflows. The new test executes the listener directly from `static/dashboard.html`, so a production refactor that drops or changes a locale re-render branch will fail the test rather than continuing to pass against a copied implementation.

This covers the locale option proposed in #284:

- the navigation, theme, language switch, fleet, dashboard data, and health views refresh when their data is available;
- switching language during startup remains safe before the three data payloads exist;
- each active tab selects only its own renderer, including the forced refresh arguments used by Experiments and Costs;
- the Disks view clears its cached row signature before repainting; and
- an exception in one renderer remains isolated so later views can still pick up the new locale.

The test changes no dashboard behavior. It only makes the existing re-render contract executable in CI.

Fixes #284.

## Why this area

The existing VM test already exercises the refresh loop, live-tail folding, in-place chart/card repainting, and stale-stream recovery in depth. Locale changes had a similarly branchy client-side path but no direct test, so extending the same source-lifting pattern there adds coverage without duplicating the refresh-loop cases.

## Checklist

- [x] **Base branch is `next`** (the integration branch), not `main`.
- [x] Branched off the **latest `next`** (`54fdb43b49241721d0776527e8bd32f415e33aae`, re-fetched immediately before commit).
- [x] Tested locally with `docker compose -p codex-hlm-284 up -d --build`; `/healthz` returned `{"status":"ok","version":"0.33.0"}` and the dashboard returned HTTP 200.
- [x] No version bump in this PR.
- [x] The new-monitor/probe pattern is not applicable; this PR adds test coverage only.

## Verification

- `node tests/js/test_locale_rerender.js` — 39/39 checks passed (including every guarded renderer as a failure source)
- `node tests/js/test_refresh_loop.js` — 80/80 checks passed
- `node tests/js/test_tflops_cells.js` — 28/28 checks passed
- `git diff --check` — passed
- Docker build, container start, `/healthz`, and dashboard HTTP smoke — passed; the test container was removed afterward

## Notes for the reviewer

The extractor deliberately looks for the listener's column-zero closing `});`; the nested `safe(() => ...)` callbacks are indented, matching the existing single-file dashboard style. If the listener is later moved or renamed, the test fails with an entry-point-specific message instead of silently exercising nothing.

AI assistance was used to help inspect candidate paths and draft the test, with the final diff reviewed and all commands above run by the contributor.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,12 @@ jobs:
       - name: Dashboard refresh-loop behaviour
         run: node tests/js/test_refresh_loop.js
 
+      # Locale changes hydrate static markup and then re-render whichever
+      # JavaScript-built view is active. Lift that listener from the shipped
+      # dashboard and cover every tab branch without needing a browser build.
+      - name: Dashboard locale re-render behaviour
+        run: node tests/js/test_locale_rerender.js
+
       # Same lift-and-run technique, for the AI Models tab's TFLOPS arithmetic.
       # The live fleet can't cover it: the hub has one card and one fully-
       # resident model, so the multi-card and RAM-spill branches only ever get
--- a/.github/workflows/release-validate.yml
+++ b/.github/workflows/release-validate.yml
@@ -67,6 +67,9 @@ jobs:
       - name: Dashboard refresh-loop behaviour
         run: node tests/js/test_refresh_loop.js
 
+      - name: Dashboard locale re-render behaviour
+        run: node tests/js/test_locale_rerender.js
+
       - name: Build the image
         run: docker build -t homelab-monitor:release-check .
 
--- a/tests/js/test_locale_rerender.js
+++ b/tests/js/test_locale_rerender.js
@@ -1,0 +1,184 @@
+#!/usr/bin/env node
+/*
+ * Locale changes hydrate static markup inside I18N.set(), then the
+ * `lang:changed` listener rebuilds the dashboard pieces whose text is produced
+ * by JavaScript. A browser smoke test can miss stale labels in tabs that were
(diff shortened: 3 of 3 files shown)


SikamikanikoBG (approved):
Verified the extracted `lang:changed` listener against `static/dashboard.html` (line 8849) — call order, per-tab renderer args, and the disks dataset-sig reset all match exactly what ships. CI wiring is a clean 3-line addition to both workflows.

PR #291, opened 2026-08-29
chore: centralize Python dependency management

### LINKED ISSUE #282: Dependencies are only pinned in the Dockerfile — add `requirements.txt` and Dependabot
The app's four runtime dependencies (`flask`, `jeepney`, `prometheus_client`, `mcp`) are pinned inline in `Dockerfile:9` and nowhere else. Consequences:

- A developer setting up a local run (see CONTRIBUTING) has to dig into the Dockerfile to learn what to `pip install`.
- Nothing tracks upstream releases for security fixes — there's no `requirements.txt` for Dependabot to watch and no `.github/dependabot.yml` at all (actions updates only, if any, are also unmanaged).

**Suggested scope**

- Extract the pins into `requirements.txt` as the single source of truth; `Dockerfile` and CI install from it.
- Add `.github/dependabot.yml` covering both `pip` and `github-actions` ecosystems.
- Update the CONTRIBUTING local-setup section to point at the file.

Mechanical, small diff, and it turns future dependency upkeep into Dependabot PRs.

xianjianlf2:
## Summary
- move runtime dependency constraints into `requirements.txt` as the shared source for Docker, CI, and local development
- add weekly Dependabot updates for pip and GitHub Actions
- document the local dependency installation command

Closes #282

## Validation
- `pip check`
- `python -m py_compile app.py probe.py`
- CI-targeted pytest suites: 77 passed
- broader `pytest tests`: 919 passed, 1 skipped; one unrelated `test_topproc` assertion assumes Linux 4KB pages and fails on this 16KB-page macOS host

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install dependencies
-        run: pip install flask pytest prometheus_client
+        run: pip install -r requirements.txt pytest
 
       - name: Run snapshot tests (determinism check)
         run: python -m pytest tests/test_snapshots.py -v --tb=short
--- a/.github/workflows/release-validate.yml
+++ b/.github/workflows/release-validate.yml
@@ -52,7 +52,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install dependencies
-        run: pip install flask pytest prometheus_client
+        run: pip install -r requirements.txt
 
       - name: VERSION / CHANGELOG / Unreleased contract
         id: release
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,13 @@ the file if you want a clean slate.
 To rebuild after a code change: `docker compose up -d --build` again. To follow
 logs: `docker compose logs -f`.
 
+For a local Python environment, install the runtime dependencies from the same
+file the Docker image and CI use:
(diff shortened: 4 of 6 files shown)


pehota (changes requested):
Thanks for this, @xianjianlf2 — nice cleanup consolidating the pins into `requirements.txt`, and the `Dockerfile` `COPY requirements.txt` reordering is a good layer-cache win too.

One gap found on review:

**`.github/workflows/release-validate.yml:55` still installs deps inline instead of from `requirements.txt`**
This job still runs `pip install flask pytest prometheus_client` (unpinned) rather than `-r requirements.txt`. Since this workflow gates the actual `v*` tag/publish with the full test suite, it means the release gate can run against different dependency versions than what ships in the image — which defeats the point of centralizing the pins here. Worth updating this job to use `requirements.txt` too.

Everything else checked out clean (pins match previous inline versions, no drift; `ci.yml`'s bare `pip install pytest` in the `smoke` job is intentionally separate per its own comment).

xianjianlf2:
Addressed the release-gate dependency drift in `93a6d84`: `.github/workflows/release-validate.yml` now installs from `requirements.txt`, matching the image and main CI path. I left the intentionally separate smoke-job install unchanged.

Validation: `python -m pytest tests/ -q` reached 918 passed / 1 skipped; two unrelated existing failures remain locally (`test_different_threads_return_different_conns` hit a SQLite lock, and `test_aggregates_workers_by_command` expected 2 MB but observed 8 MB).

pehota (approved):
Thanks, @xianjianlf2 🙌 — the `release-validate.yml:55` gap from the previous review is fixed in 93a6d84: it now installs from `requirements.txt` instead of the unpinned inline list, so the release gate and the shipped image now use the same pins. `ci.yml`'s job also switched to `-r requirements.txt` for its deps, and `requirements.txt` matches the versions previously pinned in the Dockerfile with no drift. The `Dockerfile`'s `COPY requirements.txt` + `-r` install is a nice layer-cache win too. Everything checks out — approving.

---
_Generated by [Claude Code](https://claude.ai/code)_

